PR #9, opened 2023-02-20
fix(framework): use Mediatool date strings on props to components ins…

martinmartinchan:
…tead of dateValues

### DIFF
--- a/framework/lib/components/date-picker/date-picker/date-range-picker.tsx
+++ b/framework/lib/components/date-picker/date-picker/date-range-picker.tsx
@@ -1,11 +1,12 @@
 import React, { useRef } from 'react'
-
 import { FocusScope } from '@react-aria/focus'
 import { useDateRangePickerState } from '@react-stately/datepicker'
-import { parseDate } from '@internationalized/date'
 import { useDateRangePicker } from '@react-aria/datepicker'
 import { useMultiStyleConfig } from '@chakra-ui/react'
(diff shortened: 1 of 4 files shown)


framework/lib/components/date-picker/date-picker/date-range-picker.tsx
```diff
@@ -17,23 +18,43 @@ import { InputGroup, InputRightElement } from '../../input'
 import { Icon } from '../../icon'
 import { SimpleRangeCalendar } from '../components/calendar/simple-range-calendar'
 
+const parseValue = (value: { startDate: string, endDate: string } | undefined | string) => {
+  if (isNil(value) || typeof value === 'string') return value
```
SebCodesTheWeb, line 22: You need to add an  if `isValidDateRange(value)` parseDate(value) else return null check here. Otherwise if the value is not valid it would crash the app. Null signifies empty value. You can do this as an early return instead of the one you made to check if isNil / isString

SebCodesTheWeb (changes requested):

framework/lib/components/date-picker/date-picker/date-range-picker.tsx
martinmartinchan: Fixed!

SebCodesTheWeb:
I think you need to rebase on master

SebCodesTheWeb (approved):

PR #15, opened 2023-02-21
Use nx as a task runner

TheLudd:
This sets up nx with shared cache for transpile, test and lint actions. Try
following the new instructions in the readme file.


TheLudd:
Sorry, wrong branch

PR #16, opened 2023-02-21
Use nx as a task runner

TheLudd:
This sets up nx with shared cache for transpile, test and lint actions. Try
following the new instructions in the readme file.


README.md
```diff
@@ -9,14 +9,10 @@ To start up a local sandbox environment of the Northlight UI framework:
...
-4. cd into `/tokens`
-5. run `yarn build`
-6. cd into `/icons`
-7. run `yarn build`
-8. cd into `/framework`
-9. run `yarn start`
-
-This will spin up a local environment sandbox at localhost:3008
+4. cd into `/framework`
+5. run `yarn nx demo`
```
SebCodesTheWeb, line 13: it's enough with yarn demo right?

SebCodesTheWeb (approved):

README.md
TheLudd: it is not

TheLudd: nx runs all dependant builds that is needed since you did the last build

TheLudd: You can alternatively install nx globally and run `nx demo`

PR #19, opened 2023-02-21
Upgrade dependencies in multiple steps

TheLudd:
From easy to hard, from breaking to non breaking (that actually was breaking).


SebCodesTheWeb (changes requested):
There should be over 300 test: Screenshot 2023-02-21 at 16 35 40

But when I run yarn test on this branch only a few tests run? Screenshot 2023-02-21 at 16 36 24

TheLudd:
@SebCodesTheWeb what command do you run?
If I go into framework and run `yarn test` I get 382 test passing

SebCodesTheWeb:
I run yarn test...

SebCodesTheWeb:
Me and Oles have done some evening investigation. The problem could be broken up in two parts: 1. The test failing when upgrading dependencies, 2. Mocha not reporting the error. The criminal for (1) is `framer-motion`, more specifically `v8.0.4`. We don't know what happened, bit you only change that dependency then the test fails. https://github.com/framer/motion/commits/main?after=5eac213220ce788712ea2deeb09164e4831e359d+314&branch=main&qualified_name=refs%2Fheads%2Fmain. Som of the commits that transpiled here caused the breaking change. Oles is investigating (2), whilst I continue investgating v.8.0.4 release of framer motion. Otherwise we would have to degrade framer motion to `v.8.0.3 `and it should work

TheLudd:
@SebCodesTheWeb I have removed framer-motion from the upgrades and now all tests run

SebCodesTheWeb (approved):

PR #22, opened 2023-02-22
fix(framework): fit-content on field comp instrinstic w

SebCodesTheWeb:


framework/lib/components/checkbox/checkbox-field.tsx
```diff
@@ -11,22 +12,24 @@ export const CheckboxField = ({
   validate,
   ...rest
 }: CheckboxFieldProps) => (
-  <Field
```
olchyk98, line 14: 100%

olchyk98 (changes requested):

olchyk98 (approved):

PR #24, opened 2023-02-22
fix(framework): access to form context api via ref

SebCodesTheWeb:
Previousely the form context was closed within
the <Form> tag itself, thus the form context api
was not available to methods outside of it, and
the form had to be refactored with the useForm hook to work properly.

Now a ref has been added that should act as syntax sugar to simplify this process

### DIFF
--- a/framework/lib/components/form/form.tsx
+++ b/framework/lib/components/form/form.tsx
@@ -1,16 +1,17 @@
-import React, { useEffect, useRef } from 'react'
+import React, { forwardRef, useEffect, useImperativeHandle, useRef } from 'react'
 import {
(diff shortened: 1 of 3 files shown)


framework/lib/components/form/form.tsx
```diff
@@ -38,6 +39,8 @@ export function Form<FormValues extends FieldValues> ({
       ...formSettings,
     })
 
+  useImperativeHandle(ref, always(newMethods), [])
```
magnusohlin, line 42: Nice use of this function

olchyk98 (approved):

PR #31, opened 2023-02-24
fix(framework): custom OverflowIndicator

SebCodesTheWeb:


TheLudd (approved):

framework/lib/components/tag/overflow-indicator.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import { OverflowIndicatorProps } from './types'
```
TheLudd, line 2: Note:
I do think these belong in this file and that you can import them from here in the other component. But if there is a convention in Norhtlight then let's stick with that for now.

TheLudd (approved):

PR #37, opened 2023-02-28
fix(framework): change default textDecoration

SebCodesTheWeb:
Idk why, the linkProps prop works fine in the sandbox enviroment in northlight.
But when trying in the domain sandbox the text decoration is still underline

Defaulting it to have no textDecoration should fix it, and we should probably
have it like that anyway, as when you use the clickable as a wrapper you 
don't want the text to be underlined


TheLudd (approved):

TheLudd:
published to npm

PR #40, opened 2023-03-02
feat(framework): StepStack

SebCodesTheWeb:
Screenshot 2023-03-02 at 17 40 45


framework/sandbox/docs/pages/step-stack-page/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    title="Step Stack"
+    subtitle="Step stack is a util wraper for creating ordered vertical layouts"
+  >
+    <Stack spacing={ 4 }>
+      <Stack mb={ 4 } spacing={ 4 } maxW="lg">
+        <Text>
+          Mediatool uses util <b>{ '<StepStack />' }</b>
+        </Text>
+        <StepStack maxW="sm" spacing="4">
+          { arrayWithLength(5).map((i) => (
```
TheLudd, line 24: You can do `times` instead from ramda

framework/lib/components/step-stack/step-stack.tsx
```diff
@@ -0,0 +1,56 @@
...
+  spacing = '4',
+  rowHeight = '10',
+  ...rest
+}: StepStackProps) => {
+  const rows = getChildrenWithProps(children, {})
+  const parsedRowHeight = useToken('sizes', rowHeight)
+
+  return (
+    <Stack spacing={ spacing } position="relative" { ...rest }>
+      { addIndex(map)(
```
TheLudd, line 22: `addIndex` should not be called in here, you are creating a new function on every render. It should be moved to utils as `mapIndexed`.

HOWEVER: I though we did not want to use ramda in northligth and keep dependencies to a minimum so I was a bit surprised to see it here. 
In react components we do accept `list.map(...)` since the callback function is never reused. 
Or do you prefer this style?

TheLudd (changes requested):

magnusohlin:
From the screenshot it looks like the circles are in fact not circles but ellipses. Can we get them round?

framework/sandbox/docs/pages/step-stack-page/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    title="Step Stack"
+    subtitle="Step stack is a util wraper for creating ordered vertical layouts"
+  >
+    <Stack spacing={ 4 }>
+      <Stack mb={ 4 } spacing={ 4 } maxW="lg">
+        <Text>
+          Mediatool uses util <b>{ '<StepStack />' }</b>
+        </Text>
+        <StepStack maxW="sm" spacing="4">
+          { times(identity, 5).map((i) => (
```
TheLudd, line 24: no need for identity or `.map`, just do `times(yourFn, 5)`

TheLudd (changes requested):

TheLudd (approved):

TheLudd:
Published to npm

PR #42, opened 2023-03-04
fix(framework): remove absolute imports

SebCodesTheWeb:
9140FD18-89A9-4CF3-AA6C-6A2A004BEABA


TheLudd (approved):
Nice. These things were a trick of the devil.

I took the liberty to add removal of the config that allowed them as well to this PR.

PR #48, opened 2023-03-23
fix(framework): fix the debounce of searchbar

martinmartinchan:
The old debounce did not work properly for components outside that reinitiated a new loadOptions function for every render (due to various reasons such as params used inside changing).

Doing useMemo or useCallback on the search function yielded no suggestion from the component.

This change introduces a new way of debouncing, using a timeout outside of react life cycle.

This change also removes the debounceOptions from the search bar component. Technically it is a breaking change but it is not used anywhere.


framework/lib/components/search-bar/search-bar.tsx
```diff
@@ -1,13 +1,13 @@
-import React, { forwardRef, useCallback, useState } from 'react'
+import React, { forwardRef, useMemo, useState } from 'react'
 import {
   AsyncSelect,
   GroupBase,
   InputActionMeta,
   SelectInstance,
 } from 'chakra-react-select'
 import { filter, identity, is, test, toLower } from 'ramda'
-import debounce from 'lodash.debounce'
```
TheLudd, line 9: If we remove this dependency it should be removed from package.json as well. We don't use it anywhere else right?

TheLudd (changes requested):

framework/lib/components/search-bar/search-bar.tsx
martinmartinchan: It is still used in "useSortedIcons"

framework/lib/utils/debounce/create-debounce-function-instance.ts
```diff
@@ -0,0 +1,18 @@
+/** A debounce function to be used together with Async select in SearchBar.
```
TheLudd, line 1: You wrote this? ;)

TheLudd (approved):

framework/lib/utils/debounce/create-debounce-function-instance.ts
martinmartinchan: Adjusted a stackoverflow suggestion.
Chatgpt actually only suggested debounce from lodash

PR #59, opened 2023-04-05
fix(northlight/framework): searchbar types

AnnaMalmgren:
the types for onAdd and onRemove gave ts error because of the unknown type, hence I changed it to a generic type.


SebCodesTheWeb:
This should be deprecated, check out the example "Event Meta" under select reference: https://northlight.dev/reference/select

PR #62, opened 2023-04-25
fix: remove overflow:scroll on html element

magnusohlin:
This fixes an issue with the theme where the scrollbars placeholders were always visible on the page regardless of scrolling or not.


ilija- (approved):
I'm "almost" confident about this one..

PR #68, opened 2023-05-31
chore(framework): added isClearable for range calendar

rrahman-rexhepi:
added isClearable prop for range calendar. Reason for this is that we have clear button inside calendar picker. Even though we have isClearable false or true , it was always visible.

image


magnusohlin (changes requested):
See comments above

magnusohlin (approved):

PR #69, opened 2023-06-09
chore(framework): add tooltip to avatar

unganemo:
Added the ability to display a tooltip for the Avatar component.

closes: DEV-8152

### DIFF
--- a/framework/lib/components/avatar/avatar.tsx
+++ b/framework/lib/components/avatar/avatar.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { chakra, useMultiStyleConfig } from '@chakra-ui/react'
+import { Tooltip, chakra, useMultiStyleConfig } from '@chakra-ui/react'
 import { BusinessContactDuo, UserSquareDuo } from '@northlight/icons'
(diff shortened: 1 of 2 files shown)


orestisrodriguez:
Chakra's Tooltip needs the component it wraps to `forwardRef` so it can place it correctly, but no need to add a tooltip inside the avatar itself, just the forwardRef

```
const Avatar = React.forwardRef((props, ref) => {
[...]
return (
  <HStack ref={ ref }>
    [...]
  </HStack>
)
})
```

magnusohlin:
I don't think this is the right solution to this task unfortunately. All components of Chakra are made to be able to be composed when used, this is unnecessary complication. If I want a tooltip around an Avatar I will just add a <Tooltip> around an avatar when using it. If you want a specific component that always has a tooltip one way would be to create a new Avatar component called TooltipAvatar or similar that has this functionality, be mindful of duplicated code then.

unganemo:
Ahh okay. Yeah, wrapping the Tooltip around the Avatar is how I did it first, but it ended up in the top left corner of the screen, but Orestis solution solves it.

PR #74, opened 2023-06-16
fix(components): make children prop optional in flip button

TautvydasGec:


orestisrodriguez (approved):

TautvydasGec:
> all the blind people using mediatool will be mad. but oh well .. It is what it is ..

@ilija- but we should always use aria labels no matter what. and requiring children would not help blind ppl would it?

ilija-:
I'm sorry I missed your comment. aria-labels should be used. This is part of the requirements for having compliant ( read. WCAG ) software. The only thing we aim to do is basically business value. This means that having AA standard for accessibility allows mediatool to sign contracts with more "serious" clients worldwide. You may look at it as a requirement that serious companies are asking before even negotiations start to roll.

PR #81, opened 2023-07-04
fix(framework): do not have internal state in sortable list

TautvydasGec:


unganemo (approved):
Only thing I can think of if it's being used elsewhere but fixes are on their way

PR #83, opened 2023-07-12
chore(northlight): new labelPosition prop for SwitchField

ilija-:
- Fixes the issue with the strange spacing between <Switch> and the generated <Label> 
- The new prop `labelPosition` gives an opportunity to select whether you want the label to be on the "left" or "right" of the <Switch>
- Basic usage added to the reference page

closes: DEV-8557

### DIFF
--- a/framework/lib/components/switch/switch-field.tsx
+++ b/framework/lib/components/switch/switch-field.tsx
@@ -4,12 +4,25 @@ import { SwitchFieldProps } from './types'
 import { Field } from '../form'
 import { Switch } from './switch'
 import { Box } from '../box'
+import { Flex } from '../flex'
+import { Label } from '../typography'
 
 /**
  * The switch component wrapped in a <Field />
  * meant to be used only inside <Form />
  * @see switch
  * @see {@link https://northlight.dev/reference/switch-field}
+* @example (Example)
+ * ## Basic
+ * (?
+ * <Form initialValues={{name: ''}}>
+ *  <SwitchField
+ *   name="terms"
+ *   label="I agree to the Terms & Conditions"
+ *   labelPosition="left"
+ *  />
+ * </Form>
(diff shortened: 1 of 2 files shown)


framework/lib/components/switch/switch-field.tsx
```diff
@@ -19,25 +32,43 @@ export const SwitchField = ({
...
+        )
+
+        const labelComponent = label ? <Label htmlFor={ name }>{ label }</Label> : null
+
+        return (
+          <Flex
+            gap={ 2 }
+            direction={ labelPosition === 'left' ? 'row-reverse' : 'row' }
+          >
+            { switchComponent }
```
TautvydasGec, line 67: I would just do this :arrow_down: Assigning components to variables is harder to read
`````
          <Switch
            name={ name }
            onChange={ (e) => {
              onChange(e)
              onChangeCallback(e)
            } }
            value={ value }
            data-testid="switch-field-test-id"
            { ...rest }
          />
          { label && 
                <Label htmlFor={ name }>{ label }</Label>
            }

File: framework/lib/components/switch/switch-field.tsx [hunk 2]
```diff
@@ -19,24 +32,34 @@ export const SwitchField = ({
...
-          name={ name }
-          onChange={ (e) => { onChange(e); onChangeCallback(e) } }
-          value={ value }
-          data-testid="switch-field-test-id"
-          { ...rest }
-        />
+        <Flex
+          gap={ 2 }
+          direction={ labelPosition === 'left' ? 'row-reverse' : 'row' }
+        >
```
unganemo, line 50: Does this fix so we can use justify and align the label inside the switch? Right now, the label is "stuck" in a flex-start in both justify and align

TautvydasGec (approved):

PR #84, opened 2023-07-13
chore(northlight): add icon placement prop

unganemo:
Added an icon placement prop so user can decide if icon should be placed on the left or right side of the button.
if iconPlacement is none, no icon will be shown (including checkmark).
iconPlacement defaults to left.

closes: DEV-8553


framework/lib/components/flip-button/flip-button-group.tsx
```diff
@@ -17,6 +17,7 @@ export const FlipButtonGroup = (props: FlipButtonGroupProps) => {
     children,
     direction = 'row',
     isDisabled = false,
+    hasCheckmark = true,
```
TautvydasGec, line 20: Also need it on FlipBUttonGroupField.

TautvydasGec (commented):
@ilija- what about we would add a prop like selectedIcon. we could pass 'left' | 'rigth' | 'none' ?

framework/lib/components/flip-button/flip-button-group.tsx
ilija-: This is actually not a bad idea

ilija- (approved):

TautvydasGec:
@ilija- could you check my comment?

framework/lib/components/flip-button/flip-button-group.tsx
ilija-: Sorry i missed that one. 

`showIcon: 'left' | 'rigth' | 'none' ` seems pretty understandable to me.

unganemo: @ilija- check latest
`iconPos: 'left' | 'right' | 'none'`

ilija-: Position none ?

TautvydasGec: showIcon implies action so it should be a function. @ilija-

ilija-: hmm .. then I'm okay with that. `iconPos` .. seems plausible.

ilija- (approved):

TautvydasGec (approved):

ilija-:
> @ilija- what about we would add a prop like selectedIcon. we could pass 'left' | 'rigth' | 'none' ?

Too long too much. The one above is fine IMO, but if you guys decide to go with that one I'm fine. Either way, it will serve the developers.

PR #87, opened 2023-08-08
feat(framework): add hooks for debounce, interval and loading message

JonathanNeville:
closes: DEV-8737


framework/lib/hooks/use-debounce/index.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+
+export const useDebounce = <T>(value: T, delay: number) => {
```
TautvydasGec, line 3: need tests for both of em

TautvydasGec (changes requested):

framework/lib/hooks/use-loading-message/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export const useLoadingMessage = (prefix: string = '') => {
+  const [ dots, setDots ] = useState('.')
+
+  useInterval(() => {
+    if (dots.length === 3) {
+      setDots('.')
+    } else {
+      setDots(`${dots}.`)
+    }
+  }, 300)
```
TautvydasGec, line 13: this 300 should be  a parameter we can default it to 300 but we should be able to change it

framework/test/unit/hooks/use-interval-test.ts
```diff
@@ -0,0 +1,20 @@
...
+describe('useLoadingMessage', () => {
+  it('returns the value and adds a dot to it every 300 ms', async () => {
+    const { result } = renderHook(
+      (v) => useLoadingMessage(v),
+      { initialProps: 'loading' }
+    )
+
+    assert.equal(result.current, 'loading.')
+
+    await BluebirdPromise.delay(400)
```
TautvydasGec, line 15: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor could we use this?

framework/test/unit/hooks/use-debounce-test.ts
```diff
@@ -0,0 +1,25 @@
...
+      { initialProps: 1 }
+    )
+
+    assert.equal(result.current, 1)
+
+    act(() => {
+      rerender(2)
+    })
+
+    await BluebirdPromise.delay(5)
```
TautvydasGec, line 20: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

TautvydasGec (changes requested):

PR #88, opened 2023-08-08
fix(framework): value passed to onChange of use-select-callback

lhuber92:
Fixed so that isNil is checking for val instead of value. Also added undefined val condition to second if block. Lastly, added an example of a clearable <Select> component to the docs.

### DIFF
--- a/framework/lib/hooks/use-select-callbacks/use-select-callbacks.ts
+++ b/framework/lib/hooks/use-select-callbacks/use-select-callbacks.ts
@@ -14,9 +14,9 @@ export const useSelectCallbacks = <T extends BasicOption>({
 
   const handleChange = (val: MultiValue<T> | SingleValue<T>, event: ActionMeta<T>) => {
     onChange(val, event)
-    if (!isMulti && !isNil(value)) {
+    if (!isMulti && !isNil(val)) {
       onAdd((val as T).value)
-    } else if ((val as T[]).length > (items.current as T[]).length) {
+    } else if (!isNil(val) && ((val as T[]).length > (items.current as T[]).length)) {
       onAdd(last(val as T[])?.value)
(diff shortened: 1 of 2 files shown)


framework/lib/hooks/use-select-callbacks/use-select-callbacks.ts
```diff
@@ -14,9 +14,9 @@ export const useSelectCallbacks = <T extends BasicOption>({
 
   const handleChange = (val: MultiValue<T> | SingleValue<T>, event: ActionMeta<T>) => {
     onChange(val, event)
-    if (!isMulti && !isNil(value)) {
+    if (!isMulti && !isNil(val)) {
       onAdd((val as T).value)
-    } else if ((val as T[]).length > (items.current as T[]).length) {
+    } else if (val && ((val as T[]).length > (items.current as T[]).length)) {
```
martinmartinchan, line 19: !isNil(val) here as well?

martinmartinchan (changes requested):

framework/lib/hooks/use-select-callbacks/use-select-callbacks.ts
magnusohlin: yes, isNil should be used here as well for consistency

magnusohlin (changes requested):

framework/lib/hooks/use-select-callbacks/use-select-callbacks.ts
lhuber92: Fixed :)

magnusohlin (approved):

magnusohlin (approved):

martinmartinchan (approved):

PR #91, opened 2023-08-15
chore(northlight): Omnibot tokens

HannaYankouskaya:


magnusohlin:
image

magnusohlin:
image

magnusohlin:
Glad you didn't include that bastard child of a color

tokens/lib/tokens.json
```diff
@@ -1825,6 +1825,20 @@
               "value": "{ghost}",
               "type": "color"
             }
+          },
+          "ai": {
+            "default": {
+              "value": "{color.teal.400}",
```
magnusohlin, line 1831: Why do we have teal.400 here and below? It is not used anywhere in the bot today, is it present in the sketches?

tokens/lib/tokens.json
```diff
@@ -2185,6 +2207,10 @@
           }
         }
       }
+    },
+    "ai": {
+      "value": "{color.teal.400}",
```
magnusohlin, line 2212: Same here, why not teal.500?

tokens/lib/tokens.json
HannaYankouskaya: The only place I used it was the dark theme (TOTT): we have slightly brighter colors here for all interactive elements (brand color is referring to color.blue.400 for instance)

magnusohlin: Perfect, thanks for clarifying 👍

magnusohlin (approved):

SebCodesTheWeb:
> Glad you didn't include that bastard child of a color

No way don't hate on the teal color, its my favourite, sure teal.200 might be a bastard, but it's still part of the family, and families stick together; at least that's what I told my dad but he went to buy milk anyways...

SebCodesTheWeb (changes requested):
@ilija You need to export it from components for the tokens to work, talk to me about it when you got time

PR #93, opened 2023-08-16
chore: add generics support for Field

olchyk98:
This commit implements generic types forwarding in the Field component. This is neccessary to provide an easy way to refer to a form state within a Field component callback.

Previously, with no generics support, the logic would reference "any" in "value" and "onChange":
```
<Field name="name">
  {
    ({ value: any }) => (...)
  }
</Field>
```

With the generics support the developer can pass down the form body interface to the Field component, to ensure that "value" and "onChange" have valid instance type:
```
<Field<MyFormBody> name="name">
  {
    ({ value: 'Alex' | 'Mark' }) => (...)
  }
</Field>
```

It's now also possible to pass the `control` prop to achieve the same effect:
```
<Form initialValues={ { username: 'Alice' } }>
  {
    ({ control }) => (
      <Field name="username" control={ control }>
        {
          ({ value }) => ( // "value" is "string"
            <H1>{ value }</H1>
          )
        }
      </Field>
    )
  }
</Form>
```

### DIFF
--- a/framework/lib/components/form/field.tsx
+++ b/framework/lib/components/form/field.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Controller } from 'react-hook-form'
+import { Controller, FieldPath, FieldValues } from 'react-hook-form'
 import { FormControl, FormErrorMessage, FormLabel } from '../form-control'
 import { Stack } from '../stack'
 import { FieldProps } from './types'
@@ -62,7 +62,10 @@ import { getFieldError } from '../../utils'
  * ?)
  *
  */
-export function Field ({
+export function Field<
+  FormBody extends FieldValues = FieldValues,
+  FieldName extends FieldPath<FormBody> = FieldPath<FormBody>
+> ({
   name,
   label,
   children,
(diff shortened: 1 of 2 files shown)


magnusohlin:
Looks good to me, but I'm no typescript expert, can we have @orestisrodriguez have a quick look at it as well?

olchyk98:
@magnusohlin looks like @orestisrodriguez is quite busy these days, would that be okay to invite @martinmartinchan or @TautvydasGec  to review this?

magnusohlin:
> @magnusohlin looks like @orestisrodriguez is quite busy these days, would that be okay to invite @martinmartinchan or @TautvydasGec to review this?

Added @TheLudd as reviewer instead

olchyk98:
@orestisrodriguez Added support for the "control" prop as we've discussed

framework/lib/components/form/field.tsx
```diff
@@ -61,22 +61,60 @@ import { getFieldError } from '../../utils'
...
+ * component to ensure that "value" has a valid type.
+ * The passed generic type combined with the valid
+ * "name" property ensures that "value" has the expected
+ * type received from the generic type:
+ * `<Field<MyFormBody> name="username">...</Field>`
  */
-export function Field ({
+
+export function Field<
+  FormBody extends FieldValues = FieldValues,
```
orestisrodriguez, line 101: call this FormValues for consistency with the form types

orestisrodriguez (changes requested):

framework/lib/components/form/field.tsx
olchyk98: fixed!

orestisrodriguez (approved):
i'm good with that, will let the northlight people merge

olchyk98:
@magnusohlin @ilija-

magnusohlin (approved):

PR #95, opened 2023-08-16
fix(docs): transpile reference guide on save

SebCodesTheWeb:
This adds back the previous yarn watch-docs command, renamed as transpile:watch. It rebuilds the reference docs page when one makes a difference and saves the file so one can see the changes immediately instead of having to re transpile the entire docs

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -9,6 +9,7 @@
     "dev": "yarn start",
     "build": "vite build",
     "transpile": "node src/reference/utils/generate-doc.mjs",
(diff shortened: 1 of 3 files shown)


magnusohlin:
@SebCodesTheWeb Great stuff, thanks! Could you do a branch rename please, it is not pulling down with that name for some reason.

magnusohlin:
It might be an issue as well that the PR is actually done via a fork, not sure here.

ilija-:
AMAZING !

SebCodesTheWeb:
Moved to : https://github.com/mediatool/northlight/pull/128

PR #96, opened 2023-08-17
fix(framework): change selectfieldtype

JonathanNeville:
when asserting type of event passed in callback to onchange prop in selectfield it would not be compatible with the expected type. change to account for cases of ismulti is false and true respectively

### DIFF
--- a/framework/lib/components/select/select-field.tsx
+++ b/framework/lib/components/select/select-field.tsx
@@ -10,7 +10,7 @@ import { IconButton } from '../icon-button'
 import { Icon } from '../icon'
 import { InputGroupWrapper } from '../../internal-components/input-group-wrapper/input-group-wrapper'
 
-export function SelectField<T extends Option> ({
+export function SelectField<T extends Option, K extends boolean = false> ({
   name,
   label,
   options,
@@ -23,7 +23,7 @@ export function SelectField<T extends Option> ({
   inputLeftElement,
   inputRightElement,
   ...rest
-}: SelectFieldProps<T>) {
+}: SelectFieldProps<T, K>) {
   return (
     <Field
       name={ name }
@@ -49,7 +49,7 @@ export function SelectField<T extends Option> ({
                     ? values.map((item: any) => item.value)
                     : values.value
                 )
-                onChangeCallback(values as T | T[], event)
+                onChangeCallback(values as K extends true ? T[] : T, event)
               } }
               value={
               value
--- a/framework/lib/components/select/types.ts
+++ b/framework/lib/components/select/types.ts
@@ -39,13 +39,13 @@ export interface SelectProps<T>
   customOption?: ((option: T) => JSX.Element) | null
 }
 
-export type SelectFieldProps<T> = Omit<SelectProps<T>, 'onChange'> &
-InputFieldProps & {
-  onChange?: (val: T | T[], event: ActionMeta<T>) => void
+export type SelectFieldProps<T, K extends boolean = false> = Omit<SelectProps<T>, 'onChange' | 'isMulti'> &
+Omit<InputFieldProps, 'isMulti'> & {
+  onChange?: (val: K extends true ? T[] : T, event: ActionMeta<T>) => void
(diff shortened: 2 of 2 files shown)


framework/lib/components/select/select-field.tsx
```diff
@@ -49,7 +49,7 @@ export function SelectField<T extends Option> ({
                     ? values.map((item: any) => item.value)
                     : values.value
                 )
-                onChangeCallback(values as T | T[], event)
+                onChangeCallback(values as T & T[], event)
```
olchyk98, line 52: The type intersection `T & T[]` looks like an invalid type, as `T` and `T[]` usually don't have any common methods. Could you please explain the idea behind this / provide a code sample that that is supposed to work with?

olchyk98 (changes requested):

framework/lib/components/select/select-field.tsx
olchyk98: Furthermore, if you want to change the type based on `isMulti`, you should probably use generics.

JonathanNeville: Currently if you declare e to be Option in the example below, the function is not assignable to the onChange prop because of mismatching types.
`<SelectField
                isClearable={ false }
                isMulti={ false }
                size="sm"
                aria-label="limit-dimensions-to"
                onChange={ (e: Option) => addDimensionsLimit('to', e.value) }
                value={
                  limitDimensions
                    ? {
                      value: limitPath.to,
                      label: capitalize(limitPath.to),
                    }
                    : { value: '', label: '' }
                }
                options={ [
                  { label: 'Largest', value: 'largest' },
                  { label: 'Smallest', value: 'smallest' },
                ] }
                name={ `dataSets.${name}.options.limit.to` }
                label=""
              />`

The idea is that if  isMulti is true you would change the function to be
 `
(e: Option[]) => something(map((item) => item.value, e))
`

JonathanNeville: @olchyk98  Does having the SelectFieldProps like this make more sense?

```
export type SelectFieldProps<T> = Omit<SelectProps<T>, 'onChange'> &
InputFieldProps & {
  onChange?: ((val: T, event: ActionMeta<T>) => void) | ((val: T[], event: ActionMeta<T>) => void)
  direction?: StackDirection
  name: string
  label: string
  validate?: RegisterOptions
  isRequired?: boolean
  isMulti?: boolean
}
```

olchyk98: While that may be better, this is not 100% clean. Returning "or" types in those cases requires another runtime check, otherwise you'll never know for sure if you're getting `Option[]`, or `Option`.

What I would suggest it to create a generic that would handle this behaviour, in this example generic `K`:

1. Adding another generic to the component, creates a simple declaration that can be used to link `isMulti` prop to `onChange` callback value. After a few small changes in props interface, we'll forward `K` to the props interface. 
```
export function SelectField<T extends Option, K extends boolean = true> ({
```

2. Replacing `isMulti` type from `boolean` to generic.
```
export type SelectFieldProps<T, K extends boolean = true> = ... & {
  onChange?: (val: K extends true ? T[] : T, event: ActionMeta<T>) => void
  isMulti?: K
}
``` 

Note, that we're always using `extends boolean = true`, It tells compiler that `isMulti` in that case, may be only a boolean (defaulted to true)

olchyk98 (changes requested):
Conversation: https://github.com/mediatool/northlight/pull/96/files#r1296947839

framework/lib/components/select/select-field.tsx
olchyk98: I'll be in the office soon, let me know if you need any help with this!

olchyk98 (changes requested):

olchyk98:
@JonathanNeville Sorry about the misleading information. If `isMulti` is not specified than it's `false`. Could you please default `isMulti` generics to `false`?

olchyk98 (approved):

PR #102, opened 2023-08-24
feat(framework): add outside click hook

unganemo:
Added a new outside click hook. This checks if click is inside the refs client rect or not.


framework/lib/hooks/use-outside-rect-click/index.ts
```diff
@@ -0,0 +1,31 @@
+import { isNil } from 'ramda'
```
TautvydasGec, line 1: test please. otherwise looks good

TautvydasGec (changes requested):

framework/lib/hooks/use-outside-rect-click/index.ts
unganemo: ofc it looks good, you wrote it ;)

framework/lib/hooks/use-outside-rect-click/index.ts
```diff
@@ -0,0 +1,31 @@
...
+        bottom,
+      } = ref.current.getBoundingClientRect()
+
+      const isInBound = isBetween(clientX, left, right) && isBetween(clientY, top, bottom)
+
+      if (isInBound) return
+      callback()
+    }
+
+    document.addEventListener('click', handleMouseClick)
```
magnusohlin, line 26: I think you need both click and touchstart here and when removing the listener for it to work on for example iPads `document.addEventListener('click touchstart', handleMouseClick)`

magnusohlin: For reference see Chakra-UIs implementation of useOutsideClick: https://github.com/chakra-ui/chakra-ui/blob/main/packages/hooks/use-outside-click/src/index.ts

magnusohlin (changes requested):

TautvydasGec (approved):

framework/package.json
```diff
@@ -66,7 +66,8 @@
     "react-router-dom": "5.2.0",
     "react-virtualized": "^9.22.3",
     "react-virtualized-auto-sizer": "^1.0.7",
-    "react-window": "^1.8.8"
+    "react-window": "^1.8.8",
+    "sinon": "^15.2.0"
```
magnusohlin, line 70: Added because of breaking tests

magnusohlin (approved):

PR #105, opened 2023-08-31
fix(framework): reduce zIndex for datePicker controls

olchyk98:
@ilija- @magnusohlin This is a "dirty" fix. Can potentially be merged now to fix the symptom.

---

This commit fixes the issue when DatePicker controls would overlap the multi-select dropdown. Since, the datePicker controls should be on the same layer as the input, the zIndex value was reduced to 0.


ilija-:
I think for now it's the only solution. We will need to click around ( od a quick QA ) on everything to see if we have a conflict somewhere...

ilija- (approved):

PR #108, opened 2023-09-11
feat(framework): add hook for returning users screen size

JonathanNeville:
add hook that returns a string value corresponding to the screen size of the user


TautvydasGec (approved):
this looks good to me but i would love for northlight ppl say if this what we want

framework/test/unit/hooks/use-screen-size-test.ts
```diff
@@ -0,0 +1,36 @@
...
+  dispatchEvent: (_event: Event) => true,
+})
+
+describe('useScreenSize', () => {
+  it('should return "large" for large screens', () => {
+    window.matchMedia = createMatchMedia(1280)
+    const { result } = renderHook(() => useScreenSize())
+
+    expect(result.current).to.equal('large')
+  })
```
SebCodesTheWeb, line 23: This test prob not needed, the dependency as well. If you want to keep the test then atleast move the dependency as a dev dependency

framework/lib/hooks/use-screen-size/index.ts
```diff
@@ -0,0 +1,12 @@
...
+export const useScreenSize = () => {
+  const screenSizes = useMediaQuery([ '(min-width: 1280px)', '(min-width: 1024px)' ])
+  if (screenSizes[0]) {
+    return 'large'
+  }
+  if (screenSizes[1]) {
+    return 'medium'
+  }
+  return 'small'
+}
```
SebCodesTheWeb, line 12: Use lg, md, sm instaed

SebCodesTheWeb (changes requested):

SebCodesTheWeb (approved):

PR #112, opened 2023-09-11
fix(framework): allow string based values to select

SebCodesTheWeb:
DEV-9087

https://github.com/mediatool/northlight/assets/90923099/e9215d2a-bc7d-4a47-99e1-2872c9e977c2

### DIFF
--- a/framework/lib/components/select/get-matching-value.ts
+++ b/framework/lib/components/select/get-matching-value.ts
@@ -1,0 +1,23 @@
+import { filter, find, is, length, propEq } from 'ramda'
+import { Option, SelectProps } from './types'
+
+const isStringArray = (value: any): value is string[] =>
+  is(Array, value) && length(value) > 0 && typeof value[0] === 'string'
+
+export const getMatchingValue = <T extends Option, K extends boolean>(
+  value: SelectProps<T, K>['value'],
+  options: SelectProps<T, K>['options']
+): T | T[] | undefined => {
+  if (typeof value === 'string') {
+    return find(propEq('value', value), options as T[])
+  }
+
+  if (isStringArray(value)) {
+    return filter(
+      (option) => (value).includes(option.value),
+      options as T[]
+    )
+  }
+
+  return value as K extends true ? T[] : T
+}
--- a/framework/lib/components/select/types.ts
+++ b/framework/lib/components/select/types.ts
@@ -23,8 +23,9 @@ export interface SelectProps<T, K extends boolean>
   ChakraReactSelectProps<T, K, GroupBase<T>>,
   'onChange' | 'value' | 'isMulti'
   > {
-  /** Whatever is currently selected by the select will be controlled by value prop */
-  value?: Option | Option[]
+  /** Whatever is currently selected by the select will be controlled by value prop, if value is
+   *  of type string then it will automatically pair it up with matching object from options list */
+  value?: K extends true ? T[] | string[] : T | string
   /** Take a look at the second argument, the event,
    *  for info as to which specific element was added */
   onChange?: (option: K extends true ? MultiValue<T> : SingleValue<T>, event: ActionMeta<T>) => void
--- a/framework/test/unit/components/select/get-matching-value-test.ts
+++ b/framework/test/unit/components/select/get-matching-value-test.ts
@@ -1,0 +1,31 @@
+import { expect } from 'chai'
+import { getMatchingValue } from '../../../../lib/components'
+
+describe('getMatchingValue', () => {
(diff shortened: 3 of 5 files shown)


framework/test/unit/components/select/get-matching-value-test.ts
```diff
@@ -0,0 +1,38 @@
...
+describe('getMatchingValue', () => {
+  const options = [
+    { label: 'Label1', value: 'value1' },
+    { label: 'Label2', value: 'value2' },
+    { label: 'Label3', value: 'value3' },
+  ]
+
+  it('should return the same value if it is not a string', () => {
+    const value = 42
+    // @ts-ignore
```
TautvydasGec, line 13: can we not do this? i would prefer to cast as unknown as any. instead of fully ignoring

framework/test/unit/components/select/get-matching-value-test.ts
```diff
@@ -0,0 +1,38 @@
+import { expect } from 'chai'
+import { getMatchingValue } from '../../../../lib/components' // adjust the path accordingly
```
TautvydasGec, line 2: is this comment needed?

framework/lib/components/select/get-matching-value.ts
```diff
@@ -0,0 +1,20 @@
+import { filter, find, is, length, propEq } from 'ramda'
+import { Option, SelectProps } from './types'
+
+export const getMatchingValue = <T extends Option>(
+  value: SelectProps<T>['value'],
+  options: SelectProps<T>['options']
+): T | T[] => {
+  if (typeof value === 'string') {
+    return find(propEq('value', value), options as T[]) as T
```
TautvydasGec, line 9: to not cast so many things i think we could use type guards for this. 
eg.
`export const isIncludeFilter = (filter: IncludeFilter | ExcludeFilter): filter is IncludeFilter => '$in' in filter` cause when doing if on this the later lines will know the type without needing to cast

TautvydasGec (changes requested):

framework/test/unit/components/select/get-matching-value-test.ts
SebCodesTheWeb: Oh sorry, leftover from chatgpt

TautvydasGec (approved):

framework/test/unit/components/select/get-matching-value-test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { getMatchingValue } from '../../../../lib/components'
+
+describe('getMatchingValue', () => {
+  const options = [
+    { label: 'Label1', value: 'value1' },
+    { label: 'Label2', value: 'value2' },
+    { label: 'Label3', value: 'value3' },
+  ]
+
+  it('should return the same value if it is not a string', () => {
```
olchyk98, line 11: this test is not needed, as the function does not accept `number` per its signature.

i guess, you could either refactor the function or add support for `number` :)

SebCodesTheWeb: Shit so you found the PR

SebCodesTheWeb: Yeah I just removed it

olchyk98 (approved):
LGTM

framework/lib/components/select/types.ts
```diff
@@ -20,8 +20,9 @@ export interface SelectProps<T>
   ChakraReactSelectProps<T, boolean, GroupBase<T>>,
   'onChange' | 'value'
   > {
-  /** Whatever is currently selected by the select will be controlled by value prop */
-  value?: Option | Option[]
+  /** Whatever is currently selected by the select will be controlled by value prop, if value is
+   *  of type string then it will automatically pair it up with matching object from options list */
+  value?: T | T[] | string | string[]
```
olchyk98, line 25: Is it possible to get those to be selected conditionally? (based on props)

olchyk98 (changes requested):

framework/lib/components/select/types.ts
SebCodesTheWeb: Could you elaborate on how this might be done? Not sure exactly what you have in mind, you could make another generic prop and set it equal to value prop I guess?

olchyk98: Let's discuss it when you're in the office :)

SebCodesTheWeb: @olchyk98 plz re-review this PR

olchyk98 (approved):

PR #117, opened 2023-09-12
feat(framework): rounded table variant

SebCodesTheWeb:
closed DEV-9107


framework/lib/components/table/table.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import { Table as ChakraTable } from '@chakra-ui/react'
+import { TableProps } from './types'
+
+/**
+ * @see {@link https://northlight.dev/reference/button}
```
olchyk98, line 6: `/references/table`

olchyk98:
Looks good! I kind of hate that `Th` elements are capitalized, but not sure if that's related to this PR.
Screenshot 2023-09-12 at 15 42 33

framework/lib/components/table/types.ts
```diff
@@ -0,0 +1,5 @@
+import { TableProps as ChakraTableProps } from '@chakra-ui/react'
+
+export interface TableProps extends ChakraTableProps {
```
olchyk98, line 3: It's always a good idea to `Omit`, when extending like this.
> `Omit<ChakraTableProps, 'variant'>`

olchyk98 (changes requested):
Small things, otherwise looks good!

framework/lib/components/table/index.ts
```diff
@@ -9,3 +8,4 @@ export {
   TableCaption,
   TableContainer,
 } from '@chakra-ui/react'
+export * from './table'
```
olchyk98, line 11: `TableProps` should be exported as well.
I'll cover it on my workshop soon, but the main goal is to always export props, so that they can be used without creating "magic" types (like `ComponentProps<typeof Table>`, that's just ugly.)

olchyk98 (changes requested):

olchyk98 (approved):
Beautiful!

PR #119, opened 2023-09-12
fix(framework): expose styling props for Field

olchyk98:
This commit exposes styling props for the Field
component, which allows manual and custom styling
when using the component in a real application.

This is implemented by applying non-used component props to the Stack component in use, allowing for flexible integration with other layout elements.

--
https://app.clickup.com/t/758254/DEV-9228


SebCodesTheWeb (approved):
This is fine, but I think one should wrap the field in a box/flex and pass style those containers instead of styling the field directly for keeping consistency with rest of design system

PR #121, opened 2023-09-13
fix(framework): made number input stepper arrows optional

SebCodesTheWeb:
With:
Screenshot 2023-09-13 at 09 56 44
Without
Screenshot 2023-09-13 at 09 56 05


martinmartinchan (approved):
@SebCodesTheWeb Would you check the instances in webapp or domain where we use this and see if we really want the steppers? Otherwise, this is breaking...

PR #125, opened 2023-09-13
fix(framework): added polyfill for window.matchMedia

SebCodesTheWeb:
In chakra-react-select v.4.5.0 @chakra-ui/media-query was added as a dependency. This uses window.matchMedia api which is not defined in testing environment. This polyfills should effectively stub it in the library itself and fix the issues in the tests that use Northlight

closed DEV-9240

### DIFF
--- a/framework/index.ts
+++ b/framework/index.ts
@@ -1,1 +1,3 @@
+import './matchMediaPolyfill'
+
 export * from './lib'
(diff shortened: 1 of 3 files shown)


TautvydasGec (approved):

TautvydasGec:
@SebCodesTheWeb did you test with the tests skipped in domains?

orestisrodriguez (changes requested):
matchMedia has been supported on browsers for more than 10 years, so I think this is the wrong way to fix the issue. the library doesn't need a polyfill itself, but rather the test runner's fake DOM (jsdom) environment does.

TautvydasGec:
Jest official statement is this https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom i guess you can assign that window object only in test runner

SebCodesTheWeb:
Moved to : https://github.com/mediatool/mediatool/pull/13268

PR #128, opened 2023-09-18
fix(docs): transpile reference guide on Save

SebCodesTheWeb:


TheLudd (approved):
This is not good practice.
The transpile command should take a `-w` or `--watch` flag and in that case do watching.
Then you can do `yarn transpile` or `yarn transpile -w`. 

But I will let this merge for now and we will revisit all when we get to node 18 and TS 5.

PR #132, opened 2023-09-19
fix(framework): add forward ref to select component

martinmartinchan:


framework/lib/components/select/select.tsx
```diff
@@ -149,7 +148,9 @@ export function Select<T extends Option, K extends boolean = false> ({
   value,
   icon,
   ...rest
-}: SelectProps<T, K>) {
+}: SelectProps<T, K>,
+  ref: Ref<any>
```
SebCodesTheWeb, line 152: You should type this as opposed to any, check search bar for appropiate type

framework/lib/components/select/types.ts
```diff
@@ -42,6 +42,7 @@ export interface SelectProps<T, K>
   leftIcon?: ComponentType<any>
   customOption?: ((option: T) => JSX.Element) | null
   isMulti?: K
+  ref?: Ref<any>
```
SebCodesTheWeb, line 45: Same here

SebCodesTheWeb (changes requested):

SebCodesTheWeb (approved):

PR #133, opened 2023-09-20
chore(northlight): ct level updates

HannaYankouskaya:
Semantic tokens, pressed state + some specific component tokens. A mysterious non-working borderRadius.full token should work now. Typography for headings, labels etc is added to the component level.

### DIFF
--- a/tokens/lib/tokens.json
+++ b/tokens/lib/tokens.json
@@ -1674,6 +1674,10 @@
         "value": "{alpha.2}",
         "type": "other"
       },
+      "pressed": {
+        "value": "{alpha.2}",
+        "type": "other"
+      },
       "hover-soft": {
         "value": "{alpha.3}",
         "type": "other"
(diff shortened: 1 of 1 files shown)


tokens/lib/tokens.json
```diff
@@ -4930,6 +5262,292 @@
...
+            "textCase": "{textCase.default}",
+            "textDecoration": "{textDecoration.default}"
+          },
+          "type": "typography"
+        },
+        "h2": {
+          "value": {
+            "fontFamily": "{fontFamilies.inter}",
+            "fontWeight": "{fontWeights.bold}",
+            "lineHeight": "{lineHeights.6}",
```
SebCodesTheWeb, line 5283: Hey did the headings tokens need to be moved down to component level, I think they worked as global tokens? It is causing some error when trying to transpile the tokens:

SebCodesTheWeb (changes requested):

tokens/lib/tokens.json
HannaYankouskaya: They should be used within components, so yeah I believe. I could just duplicate them probably (so they will on both levels)

SebCodesTheWeb: Reason for failing is that in this case the components typography tokens and the global typography tokens have the same name. The names should be unique, say adding st. before the names should fix it

PR #134, opened 2023-09-21
fix(framework)!: remove full width on label

SebCodesTheWeb:
closed DEV-9208

Previousely the width of the label component was set to 100%, this was confusing for developers as it behaved differently from other typography component.


unganemo (approved):
yay

SebCodesTheWeb:
I just have to double check that this does not break anything, as it does the change the width of the label component

PR #135, opened 2023-09-22
fix(framework): add missing table props

SebCodesTheWeb:


framework/lib/components/table/index.ts
```diff
@@ -8,5 +8,15 @@ export {
   TableCaption,
   TableContainer,
 } from '@chakra-ui/react'
+export type {
```
olchyk98, line 11: <3

olchyk98 (approved):

PR #137, opened 2023-09-22
chore (northlight): tokens updates

HannaYankouskaya:
Fixed naming on the component level

### DIFF
--- a/tokens/lib/tokens.json
+++ b/tokens/lib/tokens.json
@@ -1674,6 +1674,10 @@
         "value": "{alpha.2}",
         "type": "other"
       },
+      "pressed": {
+        "value": "{alpha.2}",
+        "type": "other"
+      },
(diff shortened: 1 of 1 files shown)


tokens/lib/tokens.json
```diff
@@ -3699,250 +3637,610 @@
...
+            }
+          },
+          "error": {
+            "inverted": {
+              "value": "{destructive}",
+              "type": "color"
+            },
+            "default": {
+              "value": "{st.color.bg.error.default}",
+              "type": "color"
```
SebCodesTheWeb, line 3725: The bg and border colors should not be called st.[token]... they should be bg.base, border.default etc. This is what they are called currently in app and it would otherwise be a breaking change

SebCodesTheWeb (changes requested):

PR #140, opened 2023-09-26
chore: upgrade typescript, frontend dependencies and set type module

TheLudd:
Typescript upgraded to the latest version and framework and docs are made
into es modules. When this happened a lot of type errors occured. Upgrading
all frontend dependencies solved all errors but two who needed manual correction.

This includes #139


TheLudd:
@olchyk98 @SebCodesTheWeb Can you look into the error in `framework/lib/components/form/form.tsx`?
I am not able to solve it myself.

TheLudd:
Tests should pass now.

PR #142, opened 2023-09-27
feat(northlight): make toolbox dragable

lhuber92:
https://github.com/mediatool/northlight/assets/81630176/ed70e97e-79a2-4e84-8867-8235c5a1c1d0

closes DEV-9359

### DIFF
--- a/framework/lib/components/toolbox/toolbox.tsx
+++ b/framework/lib/components/toolbox/toolbox.tsx
@@ -1,5 +1,5 @@
-import React from 'react'
-import { useMultiStyleConfig } from '@chakra-ui/system'
+import React, { useState } from 'react'
+import { useMultiStyleConfig, useToken } from '@chakra-ui/system'
 import { coreZIndex } from '@northlight/tokens'
 import { FocusScope } from '@react-aria/focus'
 import { Flex } from '../flex'
@@ -8,6 +8,9 @@ import { Box } from '../box'
 import { Portal } from '../portal'
 import { ToolboxProps } from './types'
 import { getChildrenWithProps } from '../../utils'
+
+const clamp = (minValue: number, maxValue: number, value: number) =>
+  Math.min(Math.max(value, minValue), maxValue)
 
 /**
  * Controllable Sidebar drawer
@@ -22,7 +25,7 @@ import { getChildrenWithProps } from '../../utils'
       <Button onClick={ onOpen } w="full">
         Open Toolbox
       </Button>
-      <Toolbox isOpen={ isOpen } onClose={ onClose } size="md">
+      <Toolbox isOpen={ isOpen } onClose={ onClose } size="sm">
         <ToolboxHeader>Title</ToolboxHeader>
         <ToolboxContent>
           <Stack>
@@ -41,18 +44,17 @@ import { getChildrenWithProps } from '../../utils'
  * ?)
  */
 export const Toolbox = ({
+  isResizable = false,
   isOpen = false,
   children,
   direction = 'right',
   shouldPush = false,
-  size = 'md',
-  width,
+  size = 'sm',
   onClose,
   autoFocus = true,
   ...rest
 }: ToolboxProps) => {
   const { container } = useMultiStyleConfig('Toolbox', { size })
-
   const newChildren = getChildrenWithProps(children, { onClose }, (_child, i) => i === 0)
 
   const handleKeyDown = (e: React.KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -72,13 +83,47 @@ export const Toolbox = ({
...
+                height: '100%',
+                zIndex: 2,
+                userSelect: 'none',
+              } }
+              onMouseDown={ (e) => {
+                e.preventDefault() // prevents text from being marked on drag
+                const startX = e.clientX
+                const onMouseMove = (event: { clientX: number }) => {
+                  const newWidth = adjustableWidth - (event.clientX - startX)
+                  setAdjustableWidth(Math.max(newWidth, 100)) // set minimum width to 100px
```
SebCodesTheWeb, line 115: You

SebCodesTheWeb (changes requested):

framework/lib/components/toolbox/toolbox.tsx
SebCodesTheWeb, line 115: Put a max width here as well, it should be the width of the size ="lg", the min width should also be the width of size="sm", the comment can be removed

SebCodesTheWeb (changes requested):

framework/lib/components/toolbox/toolbox.tsx
lhuber92: @SebCodesTheWeb , I have done this now, but are we sure we want it like that? As you can see, there are really a small difference between sm and lg. I think it makes the resize functionality not very usable when we can resize it in such a small range. Movie below
https://github.com/mediatool/northlight/assets/81630176/409d0d3c-02c6-4895-9e65-3557382e842c

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -62,7 +63,17 @@ export const Toolbox = ({
...
   }
 
+  const getPixelSize = (sizeToken: string) => {
+    const widthInRem = useToken('sizes', sizeToken);
+    return parseFloat(widthInRem) * 24
+  }
+
+  const defaultWidth = getPixelSize(size)
+  const minWidth = getPixelSize("sm")
+  const maxWidth = getPixelSize("lg")
```
SebCodesTheWeb, line 73: You can use 2xl instead of lg here

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -62,7 +63,17 @@ export const Toolbox = ({
     }
   }
 
+  const getPixelSize = (sizeToken: string) => {
+    const widthInRem = useToken('sizes', sizeToken);
+    return parseFloat(widthInRem) * 24
```
SebCodesTheWeb, line 68: it should be * 16 as 1 rem = 16px

SebCodesTheWeb (changes requested):

framework/lib/components/toolbox/toolbox.tsx
lhuber92: perfect!

olchyk98 (changes requested):

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -72,13 +83,49 @@ export const Toolbox = ({
...
+                position: 'absolute',
+                top: 0,
+                left: 0,
+                width: '10px',
+                height: '100%',
+                zIndex: 2,
+                userSelect: 'none',
+              } }
+              onMouseDown={ (e) => {
+                if (isResizable) {
```
olchyk98, line 111: I could not put this into a handler, because of multiple reasons:

1. It makes the `onMouseDown` handler very big (that's usually what you're trying to avoid when writing JSX)
2. It can run in a weird way on slow machines, since you "add" and "remove" handlers every time there's a mouse interaction with the object.

Could you experiment and put it into an outside handler. Something like this:
```
[isDragging, setIsDragging] = boolean

useEffect {
  bind handlers on mount
  unbind handlers on unmount
}
```

To bind handlers, ensure that you use `useRef` hook:
```
const _myRef = useRef()

...

<div ref={ _myRef }>...</div>
```

olchyk98 (changes requested):

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -72,13 +83,49 @@ export const Toolbox = ({
...
+                    document.removeEventListener('mouseup', onMouseUp)
+                  }
+                  document.addEventListener('mousemove', onMouseMove)
+                  document.addEventListener('mouseup', onMouseUp)
+                }
+              } }
+            />
             <FocusScope autoFocus={ autoFocus }>
-              { newChildren }
+              <div style={ { width: '100%', height: '100%' } }>{ newChildren }</div>
```
olchyk98, line 128: Same here. Could that be?:
```
<Box w="full" h="full">{ newChildren }</Box>
```

framework/lib/components/toolbox/types.ts
```diff
@@ -1,6 +1,7 @@
 import { FlexProps } from '@chakra-ui/react'
 
 export interface ToolboxProps extends Omit<FlexProps, 'direction'> {
+  isResizable: boolean
```
olchyk98, line 4: This is a breaking change. After bumping Northlight, all components that use `Toolbox` will fail saying that `isResizable` property has to be specified.

Making it non-optional also requires developers to specify `isResizable` in new places where `Toolbox` is used, which can be quite annoying, as it's not a common feature used with `Toolbox`. Let's mark it as optional:
```
isResizable?: boolean
```

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -72,13 +83,49 @@ export const Toolbox = ({
...
+                zIndex: 2,
+                userSelect: 'none',
+              } }
+              onMouseDown={ (e) => {
+                if (isResizable) {
+                  e.preventDefault()
+                  const startX = e.clientX
+                  const onMouseMove = (event: { clientX: number }) => {
+                    const newWidth = adjustableWidth - (event.clientX - startX)
+                    setAdjustableWidth(Math.min(Math.max(newWidth, minWidth), maxWidth))
```
olchyk98, line 116: Looks like `clamp()`... define it as a helper utility?

olchyk98 (changes requested):

framework/lib/components/toolbox/toolbox.tsx
martinmartinchan: Can you clarify here a bit? Not sure how this optimizes it as I still need the start position

SebCodesTheWeb: @olchyk98 Yes so I talked with Oles an we will leave this be, we can still move the function up like this tho: "handeMouseDown = () => ..." for readibility purposes

SebCodesTheWeb (approved):

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -40,19 +40,22 @@ import { getChildrenWithProps } from '../../utils'
  }
  * ?)
  */
+
+const clamp = (minValue: number, maxValue: number, value: number) =>
```
olchyk98, line 44: The PR is good to go, but could you please move this function before the documentation comment. Because of the placement of this function, the documentation does not render the example component :)

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -62,23 +64,68 @@ export const Toolbox = ({
...
             height: container.h as string,
-            width: toolboxWidth,
+            width: adjustableWidth,
             zIndex: coreZIndex.overlay,
           } }
         >
-          <Flex sx={ { ...container, w: toolboxWidth } } { ...rest } onKeyDown={ handleKeyDown }>
+          <Flex
+            sx={ { ...container, w: adjustableWidth } }
+            { ...rest }
```
olchyk98, line 92: This prevents rest properties from overriding `onKeyDown` and `style` props.
Please move `{...rest}` to be the last on the props list.

olchyk98 (changes requested):

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -62,23 +64,68 @@ export const Toolbox = ({
...
+            sx={ { ...container, w: adjustableWidth } }
+            onKeyDown={ handleKeyDown }
+            style={ {
+              position: 'relative',
+              overflow: 'hidden',
+            } }
+            { ...rest }
+          >
+            <Box
+              style={ {
```
olchyk98, line 100: style with native method

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -62,23 +64,68 @@ export const Toolbox = ({
...
-            width: toolboxWidth,
+            width: adjustableWidth,
             zIndex: coreZIndex.overlay,
           } }
         >
-          <Flex sx={ { ...container, w: toolboxWidth } } { ...rest } onKeyDown={ handleKeyDown }>
+          <Flex
+            sx={ { ...container, w: adjustableWidth } }
+            onKeyDown={ handleKeyDown }
+            style={ {
```
olchyk98, line 93: style with native method

olchyk98 (changes requested):

olchyk98 (approved):

PR #143, opened 2023-09-28
fix(framework): formatted number input bugs fixes

SebCodesTheWeb:
https://github.com/mediatool/northlight/assets/90923099/6c008700-4822-4184-8e8c-565d2a592908

This PR:
- enables the user to remove last number from input field
- properly sets min and max value on blur
- allows formatted number input to take an inputrightelement

### DIFF
--- a/framework/lib/components/text-field/formatted-number-input.tsx
+++ b/framework/lib/components/text-field/formatted-number-input.tsx
@@ -1,11 +1,12 @@
-import React, { ComponentType } from 'react'
-import { identity } from 'ramda'
+import React, { ComponentType, useState } from 'react'
+import { identity, isNil } from 'ramda'
 import {
   InputAttributes,
   NumberFormatValues,
   NumericFormat,
   SourceInfo,
 } from 'react-number-format'
+import { InputGroupWrapper } from '../../internal-components/input-group-wrapper/input-group-wrapper'
 import { Input } from '../input'
 import {
   FormattedNumberInputPreset,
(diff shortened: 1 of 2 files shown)


framework/lib/components/text-field/formatted-number-input.tsx
```diff
@@ -78,46 +79,65 @@ export const FormattedNumberInput = ({
   numberOfDecimals = 2,
   max = Infinity,
   min = -Infinity,
+  inputLeftElement,
+  inputRightElement,
   ...rest
 }: FormattedNumberInputProps) => {
   const props = presetMap[preset]
+  const [ v, setV ] = useState(value)
```
martinmartinchan, line 87: Why do we need a state here instead of using the value?
Seems pretty risky to shadow the value in this case

martinmartinchan (changes requested):

framework/lib/components/text-field/formatted-number-input.tsx
SebCodesTheWeb: @martinmartinchan Yes, but it is the only solution I found to be able to set/change the value when the inputed value goes above the max or below min

martinmartinchan:
@SebCodesTheWeb Did it become 100 because its the max value for the field?

TautvydasGec:
> @SebCodesTheWeb Did it become 100 because its the max value for the field?

If so feels a little weird to take over users input. maybe we should just have validation on it (lik emaking the input red or smth)? @ilija- @HannaYankouskaya what do yall think?

SebCodesTheWeb:
@TautvydasGec @martinmartinchan Yes it did become 100 because the max was set to 100, 2. ) Changing the value to max value in case the user inputs is how we currently do it with the normal number input. There is no possibility to add validation in the formatted-number-input component itself, tho I could make the border red.

framework/lib/components/text-field/formatted-number-input.tsx
```diff
@@ -78,46 +79,65 @@ export const FormattedNumberInput = ({
...
+
+  const isInputValid = () => {
```
martinmartinchan, line 89: Shouldn't this be called formatInput or something?

SebCodesTheWeb: Well its not really formatting, it is just validating that the input is within range, I renamed it to validateRange

martinmartinchan (approved):

PR #149, opened 2023-10-09
fix(framework): support types for "as" component in Button

olchyk98:
This commit fixes the issue when specifying "as" property on the button component would not lead to the base component inferring props from component specified in "as".

This code would throw an error saying that "layoutId" is not a valid property:
  <HStack as={motion.div} layoutId="a" />

This commit fixes the issue, by parsing available assume component props and appending them to the base props (ButtonProps).


TheLudd:
I'll hand the review to @SebCodesTheWeb

SebCodesTheWeb (approved):

PR #150, opened 2023-10-10
fix(framework): add optional margin to step stack circles

martinmartinchan:
Step stack circles can now have a margin as wished.


martinmartinchan:
image

SebCodesTheWeb (approved):

PR #155, opened 2023-10-12
fix(northlight/components): fix datepicker trigger alignment

ilija-:
Fixed the daterange picker vertical-alignment

closes: DEV-9365

Screenshot 2023-10-12 at 16 41 15

### DIFF
--- a/framework/lib/components/date-picker/date-picker/date-picker.tsx
+++ b/framework/lib/components/date-picker/date-picker/date-picker.tsx
@@ -108,7 +108,10 @@ export const DatePicker = (props: DatePickerProps) => {
                 />
               </Box>
             </StyledField>
-            <InputRightElement zIndex={ 0 }>
+            <InputRightElement
+              sx={ { height: '100%', paddingRight: '1' } }
+              zIndex={ 0 }
+            >
               <Trigger
                 { ...buttonProps }
                 isDisabled={ isDisabled }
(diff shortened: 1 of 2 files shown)


framework/lib/components/date-picker/components/date-field/trigger.tsx
```diff
@@ -14,7 +14,7 @@ export const Trigger = (props: TriggerProps) => {
     <Button
       { ...buttonProps }
       ref={ ref }
-      size="sm"
+      size="md"
```
SebCodesTheWeb, line 17: Change this back to sm, otherwise the bg on hover stretches out weirdly
Screenshot 2023-10-19 at 13 41 28

framework/lib/components/date-picker/date-picker/date-picker.tsx
```diff
@@ -108,7 +108,10 @@ export const DatePicker = (props: DatePickerProps) => {
                 />
               </Box>
             </StyledField>
-            <InputRightElement zIndex={ 0 }>
+            <InputRightElement
+              sx={ { height: '100%', paddingRight: '1' } }
+              zIndex={ 0 }
```
SebCodesTheWeb, line 113: Is the zIndex needed?

SebCodesTheWeb (changes requested):

framework/lib/components/date-picker/components/date-field/trigger.tsx
ilija-: well , if it's sm then it's not accessible sometimes. It's needs to look good in sm/md/lg etc.. I'm aware the background stretches a bit but this is a quickfix , we may need to give it a more attention next time, i just wanted to quickfix the bug so the users won't struggle with the overlapping elements.

magnusohlin:
The most anticipated fix of the century ⭐

magnusohlin (approved):

PR #156, opened 2023-10-19
fix(framework): changed placeholder color editabletext

SebCodesTheWeb:
https://github.com/mediatool/northlight/assets/90923099/86715332-35e7-4a93-bbe1-cf185efdcac5

@unganemo it seems to me that the placeholder actually looked like the value instead of a placeholder. I changed the color which should clear up any confusion.


unganemo (approved):
lol

PR #158, opened 2023-10-20
fix(framework): step stack

AnnaMalmgren:
when calculating the height and top for the step stack divider the spacing wasn't considered, hence the divider was positioned slightly off. I've added the spacing to both top and height.

### DIFF
--- a/framework/lib/components/step-stack/step-stack.tsx
+++ b/framework/lib/components/step-stack/step-stack.tsx
@@ -18,7 +18,7 @@ import { Box } from '../box'
   <br /><br />
   ## Simple step stack
   (?
-    <StepStack maxW="sm" spacing="4" rowHeight="10">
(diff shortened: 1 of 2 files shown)


SebCodesTheWeb:
Hey @AnnaMalmgren would you like to elaborate on this? I checkout out your branch but things kinda looked weird: 

https://github.com/mediatool/northlight/assets/90923099/c5c060b2-095a-4bac-a0a7-4b4328222571

AnnaMalmgren:
@SebCodesTheWeb hmm yes that looks weird. The problem we're having is that it looks a bit weird as it is now as well. The problem right now is that there is a gap between the vertical divider and the first icon/circle. I thought it was because we didn't take the spacing into consideration... Do you have any idea of why this problem occur? 

execution: 
Screenshot 2023-10-23 at 09 55 58

approval flow:
Screenshot 2023-10-23 at 09 56 13

PR #160, opened 2023-10-23
fix(framework): Select dropdown direction if cant fit in modal

SebCodesTheWeb:
This will make it so the select dropdown will be displayed on top of the select input in case there is not enough space at the bottom. This applies for all container, including modal.

closed: DEV-9724

https://github.com/mediatool/northlight/assets/90923099/20539948-64c0-469e-ba2d-a56263b305f0

### DIFF
--- a/framework/lib/components/select/constants.ts
+++ b/framework/lib/components/select/constants.ts
@@ -1,0 +1,4 @@
+export const DROPDOWN_MAX_HEIGHT = 300
+export const DROPDOWN_MIN_HEIGHT = 55
+export const DROPDOWN_OPTION_HEIGHT = 37
+export const DROPDOWN_PADDING = 30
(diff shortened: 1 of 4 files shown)


ilija- (approved):

magnusohlin (changes requested):
https://github.com/mediatool/northlight/assets/521837/25b379f5-4579-4c90-aa56-6e047322700d

Not exactly sure how this should work? This is when it overflows the window. Still opens below but scrolls the window. Also have you considered keyboard interactions when the select opens above the field, should the user type up-arrow then instead?

SebCodesTheWeb:
https://github.com/mediatool/northlight/assets/90923099/d7848752-b7ac-453b-98c9-dd28c7399818

@magnusohlin I have made it so that the dropdown will be displayed on top of the select input in case it does not fit into _either_ its parent div, or the viewport window. I also added a minDropdownHeightPx prop

magnusohlin (approved):

PR #174, opened 2023-11-06
feat(framework): add delete-button

SebCodesTheWeb:
https://github.com/mediatool/northlight/assets/90923099/98c37bd8-6b85-4b94-babd-e81ee8449f27


magnusohlin (changes requested):
Nice work!

The reason I wanted it to have customizable icon was to be able to use it for other actions that require intent as well, not just delete. Some small adjustments to make this even more useful:
- change name to IntentButton or something like that
- change onDelete function to onTimeout
- add customizable text for tooltip (with the default to the one you have here)
- change the aria-label of the button to a prop as well
- change name of DeleteProgressCircle to something more general

framework/lib/components/intent-button/intent-button.tsx
```diff
@@ -0,0 +1,114 @@
...
+ * )
+ *
+ * }
+ * render(<Example />)
+ *?)
+ *
+ */
+export const IntentButton = ({
+  icon = TrashDuo,
+  timeToDeleteMs = 1000,
```
magnusohlin, line 41: timeToAction or simply timeout

magnusohlin (changes requested):
one small thing left

magnusohlin (approved):

PR #177, opened 2023-11-08
fix(docs): broken avatar images

SebCodesTheWeb:


framework/lib/components/avatar/avatar-group.tsx
```diff
@@ -14,8 +14,8 @@ import { OverflowGroup } from '../overflow-group'
  * <AvatarGroup max={ 3 }>
   {[1, 2, 3, 4, 5].map((i) => (
     <Avatar
-      name="Anakin Skywalker"
-      image="src/assets/avatars/trooper.png"
+      name="Dan Abramov"
+      image="https://bit.ly/dan-abramov"
```
ilija-, line 18: Can we host these images on our server ?

ilija- (commented):
Can we host these images on our server ?

PR #196, opened 2023-11-29
fix(framework): InputLeftElement and padding

SebCodesTheWeb:
Give appropiate padding to the input element
depending on the width of InputLeftElement


unganemo:
Have you tested this in app? So stuff doesnt mess up

PR #200, opened 2023-11-30
feat(tokens): new tokens for chart color sequence

HannaYankouskaya:
Created tokens in webapp and tott for color sequence used in charts


ilija- (approved):
🚀✨ I love tokens.

PR #202, opened 2023-11-30
feat(framework): add badge and tag

Meowmir:
add file for badge in components and rework tag

Closes: DEV-10142

Screenshot 2024-01-03 at 12 04 17
Screenshot 2023-12-21 at 10 41 04

Screenshot 2024-01-03 at 12 06 17
Screenshot 2024-01-03 at 12 06 36


ilija- (changes requested):
Everything looks great. But there is a bit more tweaks we need to do, to make it 1:1 with the requested design changes.

Now we have 3 variants:

1. `solid`
2. `outline`
3. `subtle`

1 . The `solid` variant needs to be different from the solid one we have right now on this branch. 

2. The `outline` variant needs a bit of love to be the same as the component we are using right now in figma to deliver desigs to the devs.

3. The `subtle` version looks smooth 👌

Screenshot 2023-11-30 at 15 31 09

SebCodesTheWeb (approved):
Should prob get approval from Iliya also

ilija-:
> Should prob get approval from Iliya also

Running the branch right now, testing it. @SebCodesTheWeb @Meowmir  💗

framework/lib/theme/components/tag/index.ts
```diff
@@ -28,16 +25,18 @@ export const Tag: ComponentMultiStyleConfig = {
...
   },
   variants: {
-    solid: ({ bgColor, colorScheme }) => ({
+    solid: ({ theme: { colors }, colorScheme }) => ({
       container: {
-        bgColor: bgColor ?? colorScheme,
-        color: getContrastColor(bgColor ?? useToken('colors', colorScheme)),
+        bgColor: colors[colorScheme] && colors[colorScheme][500]
+          ? colors[colorScheme][500] : colorScheme,
+        color: 'white',
```
ilija-, line 32: Small change here , to be 1:1 with the design, you would need to add a simple rule re to be able to achieve that. As you can see in the [Design File](https://www.figma.com/file/0rnkmzpBhBoVl3PsbZDmGD/%E2%AC%9B%EF%B8%8F%C2%A0%C2%A0Core.mainCOMPONENTS?type=design&node-id=2-28&mode=design)

The yellow badge has a black color. Since there is no requested change for theming here, that color stays the same.

Let's add this rule in this line to make it right `color: colorScheme === 'yellow' ? 'mono.black' : 'mono.white',`

framework/lib/theme/components/tag/index.ts
```diff
@@ -28,16 +25,18 @@ export const Tag: ComponentMultiStyleConfig = {
...
       },
     }),
-    subtle: () => ({
+    subtle: ({ theme: { colors }, colorScheme }) => ({
       container: {
-        bgColor: useToken('colors', 'background.tag.subtle'),
-        color: 'text.default',
+        bgColor: colors[colorScheme] && colors[colorScheme][100]
+          ? colors[colorScheme][100] : colorScheme,
+        color: 'default',
```
ilija-, line 39: In most of the cases for text or <span> tags this would have been right, or we would use our tokens which are mapped if the theme changes. So using `text.default` which translates form black to white when the theme changes is a perfect way to handle text over layer elements.

But in this case there is no design request for text change if the theme changes. So surface + text stay the same, so in this case you will need to change this line into `color: 'mono.black',`

framework/lib/components/tag/tag.tsx
```diff
@@ -7,18 +7,50 @@ import { TagProps } from './types'
...
- *   <Tag>
-    <TagLeftIcon>
-      <Icon as={ StackSolid } />
-    </TagLeftIcon>
-    <TagLabel>Left Icon</TagLabel>
-  </Tag>
+ * +
+ * const colors = ["mediatoolBlue", "blue", "gray", "red", "green",
+ * "orange", "yellow", "teal", "purple", "pink"]
+ * const sizes = ["sm", "md", "lg"]
```
ilija-, line 17: I will let you play around with this one. Based on the [design requierements](https://www.figma.com/file/0rnkmzpBhBoVl3PsbZDmGD/%E2%AC%9B%EF%B8%8F%C2%A0%C2%A0Core.mainCOMPONENTS?type=design&node-id=2-28&mode=design)

The `size="sm"` tag uses icon `size="xs"`
BUT
The `size="md"` tag uses icon `size="md"`
and The` size="lg"` tag uses icon `size="lg"`

Let's show this properly in the example. This is very important for the **Developer Experience** because the Developers will adopt the way that is served in the documentation, and these finesses or aesthetic models will be just copied/pasted in the next use of the tag.

ilija- (changes requested):
Great work, just a small changes to make it 1:1 with the design. ✨

Meowmir:
I added a comment of the size to the default tag.

framework/lib/theme/components/badge/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+  },
+  variants: {
+    solid: ({ colorScheme, theme: { colors } }) => ({
+      bgColor: colors[colorScheme] && colors[colorScheme][500]
+        ? colors[colorScheme][500] : colorScheme,
+      color: 'white',
+    }),
+    outline: ({ colorScheme, theme: { colors } }) => ({
+      bgColor: 'white',
```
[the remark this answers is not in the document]
ilija-: Let's clean up this code a bit. This is what i came up with, also updated in figma.

```
outline: ({ colorScheme, theme: { colors } }) => {
      const textColor = colorScheme === 'mediatoolBlue'
        ? colors[colorScheme][500]
        : colors[colorScheme] && colors[colorScheme][700]

      return {
        bgColor: 'mono.transparent',
        color: textColor,
      }
    },
```

Here is the outcome:
the 'mediatoolBlue' has a temporary color of [500], but we will update that in the future for sure. Experiencing some problems with switching the theme and inheriting color contrast atm.

Screenshot 2023-12-22 at 20 24 04

framework/lib/components/tag/tag.tsx
```diff
@@ -7,18 +7,69 @@ import { TagProps } from './types'
  * @see {@link https://northlight.dev/reference/tag}
  *
  * @example
- * (? <Tag>Brand 1</Tag> ?)
+ * Default size medium.
```
ilija-, line 10: Please add a proper documentation with heading ## and and explanation, here and let's make it a bit more professional than just a few words. Remove the "default size is medium"
Screenshot 2023-12-22 at 20 26 25

Something like

Tag
Tag component is used for items that need to be labeled, categorized, or organized using keywords that describe them.

ilija- (changes requested):
Small `<Badge>` changes and I think we are ready to release this one 🚀

ilija- (approved):
Ready to GO 🚀

PR #209, opened 2023-12-06
chore(framework): fix colors and example presentation for status pin

Meowmir:
change colors to have more contrast to background, add new variants and add all variants to example presentation in northlight

Closes: DEV-10293

Screenshot 2023-12-06 at 11 22 57

### DIFF
--- a/framework/lib/components/pin-input/types.ts
+++ b/framework/lib/components/pin-input/types.ts
@@ -1,6 +1,6 @@
 import { PinInputProps as ChakraPinInputProps } from '@chakra-ui/react'
 
-export type PinVariant = 'green' | 'yellow' | 'gray' | 'red'
+export type PinVariant = 'green' | 'running' | 'yellow' | 'inProgress' | 'gray' | 'notExecuted' | 'red' | 'rejected'
 export type PinSize = 'sm' | 'md' | 'lg'
 export type PinSizeTuple = [ number, number ]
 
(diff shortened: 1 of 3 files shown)


framework/lib/components/pin-input/types.ts
```diff
@@ -1,6 +1,6 @@
 import { PinInputProps as ChakraPinInputProps } from '@chakra-ui/react'
 
-export type PinVariant = 'green' | 'yellow' | 'gray' | 'red'
+export type PinVariant = 'running' | 'inProgress' | 'notExecuted' | 'rejected'
 export type PinSize = 'sm' | 'md' | 'lg'
```
SebCodesTheWeb, line 4: Why have the names of the variants been changed? this would be a breaking change on multiple places in domains. If we choose to go with variable names then please make the PR as a breaking change and make sure that it it bumped and fixed in those places where the old variant names are used.

SebCodesTheWeb (changes requested):

framework/lib/components/pin-input/types.ts
TheLudd: It is also possible to keep the old and add the new at the same time. 
That way, it is not a breaking change anymore and we can migrate without a major new version.

Meowmir: I have added both old and new variant names

SebCodesTheWeb (approved):

PR #210, opened 2023-12-06
feat(framework): add ai variants to respective component

SebCodesTheWeb:
This should all the ai variants to respective components as mentioned in this sketch: https://www.figma.com/file/jL2ufBP6ljlj5pcFv8HyYU/%F0%9F%93%81--Pages.AIIntegrations?type=design&node-id=1499-133096&mode=design&t=0gBjpoguMVwEpPCa-0


framework/lib/components/file-picker/types.ts
```diff
@@ -50,6 +50,7 @@ export interface FilePickerProps {
   maxFileSize?: number
   value?: string | null
   loadWithSpinner?: boolean
+  variant?: 'ai'
```
magnusohlin, line 53: Do we want to default the variant to ai? Same for below in this file

SebCodesTheWeb: No,  the default is undefined which in that case is just baseStyle.

PR #219, opened 2023-12-18
fix(framework): do not override icon size

SebCodesTheWeb:
previousely you could not set a custom size for the icon in an iconbutton as it would be overriden, this should clear up confusion, so if a size for the icon is specified that will be used, and if not then the size of the icon button will be used instead


lhuber92 (approved):
LGTM 👍

PR #222, opened 2023-12-20
fix(framework): data-testid suppport for all components

SebCodesTheWeb:


olchyk98 (approved):
Actual king @SebCodesTheWeb

PR #227, opened 2023-12-27
fix(northlight): slideFade hideDisplay default state

ilija-:
Set hideDisplay default state to `true`

closes: DEV-10342

* [Requested](https://app.clickup.com/t/758254/DEV-10342) by @JonathanNeville .
```If prop 'hideDisplay' is not passed with value true. The slideFade will prevent users from clicking area under the slideFade even if the slide is closed. I think that there are not many cases when this behaviour is desired and that it's easy to miss passing this prop to the component.```


SebCodesTheWeb:
@ilija- This is a breaking change, make sure to write that in the commit and also check in webapp that all the slideFades work as expected with hideDisplay set to true

ilija-:
> @ilija- This is a breaking change, make sure to write that in the commit and also check in webapp that all the slideFades work as expected with hideDisplay set to true

Will do !

PR #230, opened 2023-12-29
chore (northlight): new color tokens for text

HannaYankouskaya:
Added an alternate set of tokens for text on a coloured background ('over-alt') in web app and tott skins. Created specific component tokens for text in button's variants. Added standard brand, success and danger tokens for text on a white background


ilija- (approved):
Great. Now let's test them out : )

PR #232, opened 2024-01-02
fix(framework): add custom test-id to select menu component

SebCodesTheWeb:
Due to recent changes, the select menu is rendered in a portal this means it needs a separate testid from the select itself


unganemo (approved):
LGTM

PR #233, opened 2024-01-02
fix(framework): fix multi sort infinite re-render bug

SebCodesTheWeb:
Previousely the useEffect which dependend on items caused an infinite render since it updated the items state indirectly which itself depended on


unganemo (approved):
I was gonna say "have you tested this so it doesnt break anything in the app" but then I realized we dont use this anywhere in the app hehe

PR #235, opened 2024-01-04
fix(framework): fix broken token colors

SebCodesTheWeb:
This PR fixes two recent bugs:

1. Random blue color – the default button variant was recently changed to have color set to blue.500, since a lot of components, especially text and icons are wrapped in a <Button /> this made them blue

2. Badge and Tag weird colors. The text color of all badge and tag children were set to white unless colorScheme was set to yellow. Using getContrastColor together with useToken instead ensures that the correct text color is used

E6D4BF10-3B1D-44A0-859E-FACF916DAEBD
0097477E-95F4-4F16-AD16-40F3D38F5D32_4_5005_c
46E29FA3-F06D-4B16-8648-3E85249ADA7B_4_5005_c


TautvydasGec (approved):
:clap: Thanks!

PR #245, opened 2024-01-19
chore (northlight): added semantic ai-alt color token

HannaYankouskaya:
Added semantic ai-alt color tokens to webapp and tott. They are going to be used as bg colors in ai-variant of toast/alert/tooltip


HannaYankouskaya:
Correction: it is to use in toast/alert

ilija- (approved):
lgtm

PR #246, opened 2024-01-19
fix(framework): date-range-picker-fixes

SebCodesTheWeb:
closed: DEV-10597


orestisrodriguez (changes requested):
can you split into multiple commits, one for each change and add a small description to why the change was made?

orestisrodriguez (approved):

PR #250, opened 2024-02-01
fix(framework): add jsx to tooltip type

lhuber92:
changed tooltip type to accept JSX and also added an example
Screenshot 2024-02-01 at 14 33 54


framework/lib/components/tooltip/types.ts
```diff
@@ -4,6 +4,6 @@ export type TooltipVariants = 'success' | 'warning' | 'error' | 'info' | 'danger
 
 export interface OurTooltipProps extends ChakraTooltipProps {
   variant?: TooltipVariants
-  description?: string
+  description?: React.ReactNode
```
olchyk98, line 7: Good choice there!

olchyk98 (approved):

PR #252, opened 2024-02-12
feat(lib): optional currentoption prop for createable select

lhuber92:
there is now a currentOption prop that will set the currently selected option for the createable select component.

### DIFF
--- a/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
+++ b/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
@@ -1,8 +1,9 @@
-import React, { useState } from 'react'
-import { CreatableSelect } from 'chakra-react-select'
-import type { Option } from '@northlight/ui'
+import React, { useMemo, useState } from 'react'
+import { ActionMeta, CreatableSelect } from 'chakra-react-select'
+import type { Option, SingleValue } from '@northlight/ui'
 import { PlusSolid } from '@northlight/icons'
 import { Icon } from '@chakra-ui/react'
+import { customSelectStyles } from '../../theme/components/select/custom-select'
 import type { CreatableSelectDropdownProps, CreationOption } from './types'
 import { customComponents } from './custom-components'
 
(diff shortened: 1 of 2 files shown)


framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -64,6 +96,8 @@ export const CreatableSelectDropdown = ({
     setNewOptionText(newValue)
   }
 
+  console.log('HHHHHE')
```
TheLudd, line 99: :bomb:

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -54,8 +85,9 @@ export const CreatableSelectDropdown = ({
   onOptionChange,
   width = '100%',
   variant = 'outline',
+  currentOption = null,
```
TheLudd, line 88: Not sure if we have a naming convention here.
Like `initialSelection` or something? do you know @SebCodesTheWeb ?

TheLudd (changes requested):

README.md
```diff
@@ -11,6 +11,7 @@ To start up a local environment of the Northlight UI framework docs:
 3) Run `yarn`.
 4) Open one teminal, navigate to the /docs directory, and run `yarn nx dev`.
 5) Open another terminal, navigate to the /framework directory, and `yarn watch`.
+6) Open a third terminal, navigate to the /docs directory and run `yarn transpile:watch`
```
olchyk98, line 14: ??

framework/lib/components/createable-select-dropdown/types.ts
```diff
@@ -51,4 +51,9 @@ export interface CreatableSelectDropdownProps {
    *
   */
   variant?: Variant
+
+  /**
+   * The currently selected option. Useful for initializing the component with an option selected.
```
olchyk98, line 56: `currently selected option` indicates that the property can control the current state of the select. That is not true.

As @TheLudd pointed out, this should be called `initialValue`

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
olchyk98: This does not work then. When you take property and use it as default value for `useState`, if the property changes the state value won't be updated.

`const [state, setState ] = useState(defaultValue, ...otherArgs)`

---
```
const { prop1 } = props
const [ state, setState ] = useState(prop1, ...otherArgs) // << prop1 is default state value

setState(123) // << state updated, if you change prop1 from outside, nothing will change here
```

README.md
lhuber92: We have to do this, unless we cant to re-transpile manually to update the comments that forms the examples

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { useState } from 'react'
-import { CreatableSelect } from 'chakra-react-select'
-import type { Option } from '@northlight/ui'
+import { ActionMeta, CreatableSelect } from 'chakra-react-select'
+import type { Option, SingleValue } from '@northlight/ui'
 import { PlusSolid } from '@northlight/icons'
 import { Icon } from '@chakra-ui/react'
+import { customSelectStyles } from '../../theme/components/select/custom-select'
```
olchyk98, line 6: ??

olchyk98 (changes requested):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
lhuber92: You're right 👍

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
lhuber92: this is to add the same styling to createable select as normal select, such as blue dropdown icon to open it

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -105,6 +144,12 @@ export const CreatableSelectDropdown = ({
     }
   }
 
+  useEffect(() => {
+    if (initialValue) {
+      setSelectedOption(standardOptions.find((option) => option.value === initialValue) || null)
+    }
+  }, [ initialValue ])
```
olchyk98, line 151: I don't really understand what you're trying to achieve in this PR. I am pretty sure `Select` has prop `value`, which can be used to control the current state.

Name `initialValue` indicates to the developer that the value will be used only one time. And once the select component is rendered, changing the prop value from the parent component won't affect the current state.

olchyk98 (changes requested):

olchyk98:
@SebCodesTheWeb Could you check this?

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
lhuber92: @olchyk98 , we need to be able to set an initial value, hence this PR and the name :) Its not required to control that value from the outside. We use CreatableSelect underneath.

SebCodesTheWeb: (I.e. you should have two props, one called initialValue and one just called value)

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -105,6 +144,12 @@ export const CreatableSelectDropdown = ({
     }
   }
 
+  useEffect(() => {
+    if (initialValue) {
+      setSelectedOption(standardOptions.find((option) => option.value === initialValue) || null)
```
TheLudd, line 149: Don't use a useEffect for this. Set this as the initial state
You can use `useMemo` if you dont want it to search on every render

TheLudd (changes requested):

TheLudd (approved):

olchyk98 (approved):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -54,8 +88,13 @@ export const CreatableSelectDropdown = ({
   onOptionChange,
   width = '100%',
   variant = 'outline',
+  initialValue = '',
 }: CreatableSelectDropdownProps) => {
-  const [ selectedOption, setSelectedOption ] = useState<Option | null>(null)
+  const initialSelectedOption = useMemo(() => standardOptions.find(
```
olchyk98, line 93: Well done on fixing the useEffect part.
One problem with this is it's going to iterate over `standardOptions` every time `standardOptions` change, which is not really neccessary.

My suggestion is that you remove the dependencies list, so that this logic would be triggered only once (upon mounting).

olchyk98 (changes requested):

olchyk98 (approved):

PR #266, opened 2024-03-08
feat(framework-lib): made selection controllable from outside

lhuber92:
createableSelectDropdown is now controllable from the outside using its value prop.

closes DEV-12261


framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -89,15 +90,15 @@ export const CreatableSelectDropdown = <T extends string = string> ({
   onOptionChange,
   width = '100%',
   variant = 'outline',
-  initialValue,
```
olchyk98, line 92: we still want to support `initialValue`.
Check if it's implemented in `<Select />`

olchyk98 (changes requested):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
SebCodesTheWeb: I think it's called defaultValue for select

olchyk98: So tru!

olchyk98:
@SebCodesTheWeb take a lead :)

olchyk98 (approved):
@SebCodesTheWeb is in charge

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -89,15 +90,18 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
+  value,
   menuPlacement = 'bottom',
 }: CreatableSelectDropdownProps<T>) => {
   const initialSelectedOption = useMemo(() => {
-    if (isNil(initialValue)) {
-      return null
+    if (!isNil(value)) {
+      return standardOptions.find((option) => option.value === value) ?? null
+    } if (!isNil(defaultValue)) {
+      return standardOptions.find((option) => option.value === defaultValue) ?? null
```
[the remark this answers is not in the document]
lhuber92: Standardoptions needs to have a value prop, it is of the Option type.

I cannot set the initialSelectedOption as only the value like your image, value needs to be of the type Option, that is used internally to find the correct option, that is how Oles wanted it to be.

PR #267, opened 2024-03-08
fix(framework): add striped table variant

SebCodesTheWeb:
With the addition of the custom rounded table variant, it seems to have overriddden chakra-uis inbuilt stripped variant. This variant has been copy pasted from the source code


framework/lib/theme/components/table/index.ts
```diff
@@ -25,5 +35,43 @@ export const Table: ComponentMultiStyleConfig = {
         borderTopWidth: 'xs',
       },
     }),
+    striped: (props) => {
+      const { colorScheme: c } = props
+
+      return {
+        columnHeader: {
+          color: mode('gray.600', 'gray.400')(props),
```
orestisrodriguez, line 43: this should probably be text.subdued, no ?

framework/lib/theme/components/table/index.ts
```diff
@@ -25,5 +35,43 @@ export const Table: ComponentMultiStyleConfig = {
         borderTopWidth: 'xs',
       },
     }),
+    striped: (props) => {
+      const { colorScheme: c } = props
```
orestisrodriguez, line 39: can we make sure colorScheme is defaulted?

framework/lib/theme/components/table/index.ts
```diff
@@ -25,5 +35,43 @@ export const Table: ComponentMultiStyleConfig = {
...
+
+      return {
+        columnHeader: {
+          color: mode('gray.600', 'gray.400')(props),
+          borderBottomWidth: '1px',
```
orestisrodriguez, line 44: same here, can we have the same borders as the other table variants?

orestisrodriguez (approved):

orestisrodriguez (changes requested):
see above

SebCodesTheWeb:
@orestisrodriguez pls review this PR sometime this month

framework/lib/theme/components/table/index.ts
```diff
@@ -25,5 +35,47 @@ export const Table: ComponentMultiStyleConfig = {
         borderTopWidth: 'xs',
       },
     }),
+    striped: (props) => {
+      const { colorScheme: c = 'gray' } = props
```
ilija-, line 39: Can we execute a new token for this one? Since it will be nice to have "themed" stripes in the future. So a component token for the table can be changed if changing theme in the future. What do you think?

orestisrodriguez (approved):

PR #270, opened 2024-03-21
fix(framework): createableselectdropdown has new add new field option

GustafGronlund:
The CreatableSelectDropdown component has been updated to improve its functionality. When a user selects to "Add Field," the element now transitions from an Option to EditableText, streamlining the user experience. This enhancement allows for a more seamless process in adding fields, complete with appropriate 'Submit' and 'Cancel' buttons. Additionally, a new variant has been introduced, enabling color customization of the component.

closes: DEV-12420

### DIFF
--- a/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
+++ b/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
@@ -1,12 +1,13 @@
-import React, { useMemo, useRef, useState } from 'react'
-import { ActionMeta, CreatableSelect } from 'chakra-react-select'
-import type { Option, SingleValue } from '@northlight/ui'
+import React, { useRef, useState } from 'react'
+import { ActionMeta, CreatableSelect, SingleValue } from 'chakra-react-select'
 import { PlusSolid } from '@northlight/icons'
-import { Box, Icon, useOutsideClick } from '@chakra-ui/react'
-import { isNil } from 'ramda'
+import { Box, Icon } from '@chakra-ui/react'
+import { any, isEmpty, toLower, trim } from 'ramda'
 import { customSelectStyles } from '../../theme/components/select/custom-select'
 import type { CreatableSelectDropdownProps, CreationOption } from './types'
 import { customComponents } from './custom-components'
+import { EditableText } from '../editable-text'
+import type { Option } from '../select/types'
 
 /**
  * A dropdown component that allows users to select a value from given options
@@ -23,7 +24,7 @@ import { customComponents } from './custom-components'
  *     { label: 'Snoop Doggy Dogg', value: 'snoop-dogg' },
  *   ];
  *
-*   const someOtherOptions = [
+ *   const someOtherOptions = [
  *     { label: 'Vision', value: 'vision' },
  *     { label: 'Technique', value: 'technique' },
  *     { label: 'Expression', value: 'expression' }
@@ -78,95 +79,102 @@ import { customComponents } from './custom-components'
  * ?)
  */
 
-export const CreatableSelectDropdown = <T extends string = string> ({
+const CREATION_OPTION_VALUE = 'add_field'
+
+export const CreatableSelectDropdown = <T extends string = string>({
   standardOptions,
   initialPlaceholder = 'Select or create...',
-  addOptionPlaceholder = 'Enter text...',
+  addOptionPlaceholder = 'Select or create...',
   creationOption = {
-    label: 'Add option...',
-    value: 'Add option...',
+    label: 'Add field',
+    value: CREATION_OPTION_VALUE,
     isCreation: true,
   },
   onOptionChange,
-  width = '100%',
   variant = 'outline',
   defaultValue,
   value,
   menuPlacement = 'bottom',
 }: CreatableSelectDropdownProps<T>) => {
-  const initialSelectedOption = useMemo(() => {
-    if (!isNil(value)) {
-      return standardOptions.find((option) => option.value === value) ?? null
-    } if (!isNil(defaultValue)) {
-      return standardOptions.find((option) => option.value === defaultValue) ?? null
-    }
-    return null
-  }, [ value ]
-  )
+  const [ selectedOption, setSelectedOption ] = useState<Option | null>(() => {
+    const targetValue = value ?? defaultValue ?? null
+    return targetValue
+      ? standardOptions.find((option) => option.value === targetValue) ?? null
+      : null
+  })
 
-  const [ selectedOption, setSelectedOption ] =
-    useState<Option | null>(initialSelectedOption)
-  const [ newOptionText, setNewOptionText ] = useState('')
-  const [ newOptionPlaceholder, setNewOptionPlaceholder ] = useState(initialPlaceholder)
+  const [ newOptionPlaceholder, setNewOptionPlaceholder ] =
+    useState(initialPlaceholder)
   const [ createdOptions, setCreatedOptions ] = useState<Option[]>([])
+  const [ createNewOption, setCreateNewOption ] = useState<boolean>(false)
+  const [ addFieldInputValue, setAddFieldInputValue ] = useState<string>('')
 
   function isCreationOption (option: any): option is CreationOption {
     return option && typeof option.isCreation === 'boolean'
   }
 
   const ref = useRef<HTMLDivElement>(null)
-  useOutsideClick({
-    ref,
-    handler: () => {
-      if (isCreationOption(selectedOption)) {
-        setSelectedOption(initialSelectedOption)
-        setNewOptionPlaceholder(initialPlaceholder)
+
+  const checkIfNewOptionTextExistsInOptions = (
+    string: string,
+    options: Option[]
+  ) => any((option) => toLower(option.value) === string, options)
+
+  const handleCreateOption = (newOptionTextString: string) => {
+    setCreateNewOption(true)
+
+    if (isEmpty(trim(newOptionTextString))) {
+      return
+    }
+
+    const newOptionTextLower = toLower(newOptionTextString)
+
+    const newOptionExistsInStandardOptions =
+      checkIfNewOptionTextExistsInOptions(newOptionTextLower, standardOptions)
+
+    const newOptionExistsInCreatedOptions = checkIfNewOptionTextExistsInOptions(
+      newOptionTextLower,
+      createdOptions
+    )
+
+    if (!newOptionExistsInCreatedOptions && !newOptionExistsInStandardOptions) {
+      const newOption = {
+        label: newOptionTextString,
+        value: newOptionTextString,
       }
-    },
-  })
-
-  const handleInputChange = (newValue: string) => {
-    setNewOptionText(newValue)
-  }
-
-  const handleCreateOption = () => {
-    const newOptionTextLower = newOptionText.toLowerCase()
-    const newOptionExists = standardOptions.some((option) =>
-      option.value.toLowerCase() === newOptionTextLower)
-
-    if (!newOptionExists) {
-      const newOption = { label: newOptionText, value: newOptionText }
-      setCreatedOptions((currentCreatedOptions) => [ ...currentCreatedOptions, newOption ])
+      setCreatedOptions((currentCreatedOptions) => [
+        ...currentCreatedOptions,
+        newOption,
+      ])
       setSelectedOption(newOption)
       onOptionChange(newOption)
     }
-
-    setNewOptionText('')
     setNewOptionPlaceholder(initialPlaceholder)
   }
 
   const handleChange = (
     newValue: SingleValue<Option>,
-    _actionMeta: ActionMeta<Option>) => {
+    _actionMeta: ActionMeta<Option>
+  ) => {
     const option = newValue
 
-    if (option === null) {
+    if (newValue?.isCreation) {
+      setCreateNewOption(true)
+    }
+
+    if (option == null) {
       return
     }
+
     if (isCreationOption(option)) {
       setNewOptionPlaceholder(addOptionPlaceholder)
     } else {
       setNewOptionPlaceholder(initialPlaceholder)
     }
-    setSelectedOption(option)
-    onOptionChange(option)
-  }
-
-  const handleKeyDown = (event: { key: string, preventDefault: () => void }) => {
-    if (event.key === 'Enter' && newOptionText) {
-      handleCreateOption()
-      event.preventDefault()
+    if (option.value !== CREATION_OPTION_VALUE) {
+      setSelectedOption(option)
     }
+    onOptionChange(option)
   }
 
   const combinedOptions = [ ...standardOptions, ...createdOptions ]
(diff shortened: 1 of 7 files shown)


framework/lib/components/createable-select-dropdown/custom-components.tsx
```diff
@@ -4,20 +4,40 @@ import { PlusSolid } from '@northlight/icons'
...
+              justifyContent="center"
+              alignItems="center"
+            >
+              <Icon as={ PlusSolid } color="brand" />
+            </Flex>
+          ) : (
+            <Box mr={ 3 } width={ 1.5 } />
+          ) }
+          <Text
+            color={ props.data.isCreation ? 'brand' : 'default' }
```
ilija-, line 34: Sorry for popping up uninvited. Can we use the system tokens here instead of the reference tokens:
`'text.brand'`
`'text.default'`

These tokens reflect decisions and describe intention/usage/quality. Please aim to use only system tokens. DM me for further explanation.

ilija- (approved):
lgtm

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -125,29 +135,31 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
-
-  const handleCreateOption = () => {
-    const newOptionTextLower = newOptionText.toLowerCase()
-    const newOptionExists = standardOptions.some((option) =>
-      option.value.toLowerCase() === newOptionTextLower)
+  const handleCreateOption = (newOptionTextString: string = '') => {
+    const newOptionTextLower = newOptionTextString.toLowerCase()
+    const newOptionExists = standardOptions.some(
+      (option) => option.value.toLowerCase() === newOptionTextLower
+    )
```
SebCodesTheWeb, line 142: I think this bug was here before, but you want to check if the newOptionTextString is included in `createdOptions` array as well, otherwise you could create the same option twice

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -97,24 +97,34 @@ export const CreatableSelectDropdown = <T extends string = string> ({
   const initialSelectedOption = useMemo(() => {
     if (!isNil(value)) {
       return standardOptions.find((option) => option.value === value) ?? null
-    } if (!isNil(defaultValue)) {
-      return standardOptions.find((option) => option.value === defaultValue) ?? null
+    }
+    if (!isNil(defaultValue)) {
+      return (
+        standardOptions.find((option) => option.value === defaultValue) ?? null
+      )
```
olchyk98, line 104: don't really like this duplication.
could we maybe define a function for that, where you pass in `value`/`defaultValue`?

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -125,29 +135,31 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
 
-  const handleInputChange = (newValue: string) => {
-    setNewOptionText(newValue)
-  }
-
-  const handleCreateOption = () => {
-    const newOptionTextLower = newOptionText.toLowerCase()
-    const newOptionExists = standardOptions.some((option) =>
-      option.value.toLowerCase() === newOptionTextLower)
+  const handleCreateOption = (newOptionTextString: string = '') => {
```
olchyk98, line 138: no need for :string here

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -125,29 +135,31 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
+    const newOptionTextLower = newOptionTextString.toLowerCase()
+    const newOptionExists = standardOptions.some(
+      (option) => option.value.toLowerCase() === newOptionTextLower
```
olchyk98, line 141: use `toLower` from ramda

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -125,29 +135,31 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
   }
 
   const handleChange = (
     newValue: SingleValue<Option>,
-    _actionMeta: ActionMeta<Option>) => {
+    _actionMeta: ActionMeta<Option>
+  ) => {
     const option = newValue
 
     if (option === null) {
```
SebCodesTheWeb, line 165: Also, one thing that has been there before but would need fixing. If you try selecting add field by pressing space or enter when focused on that option, it will actually select 'add field'. Check in the handleChange if `newValue` is equal to Add field, and toggleCreateOption then instead

SebCodesTheWeb (changes requested):

framework/lib/components/createable-select-dropdown/custom-components.tsx
```diff
@@ -4,20 +4,45 @@ import { PlusSolid } from '@northlight/icons'
...
+          </Text>
+        </>
+      </chakraComponents.Option>
+    </Box>
   ),
+  // Input: ({ children, ...props }: any) => (
+  //   <chakraComponents.Input { ...props } readOnly={ true }>
+  //     { children }
+  //   </chakraComponents.Input>
+  // ),
```
olchyk98, line 47: no comments

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -180,36 +185,59 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
-            ...(isSelected && {
-              color: 'black',
+    <Box ref={ ref } w="sm" maxW="full">
+      { createNewOption ? (
+        <EditableText
+          value={ addFieldInputValue }
+          onChange={ (v) => { setAddFieldInputValue(v) } }
+          startWithEditView={ true }
+          submitOnBlur={ false }
+          leftItem={ undefined }
```
olchyk98, line 195: not needed

olchyk98 (changes requested):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
SebCodesTheWeb: submitOnBlur and startWithEdtiView is needed

olchyk98: I meant `leftItem`

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -78,95 +79,98 @@ import { customComponents } from './custom-components'
...
-  const [ newOptionPlaceholder, setNewOptionPlaceholder ] = useState(initialPlaceholder)
+  const [ newOptionPlaceholder, setNewOptionPlaceholder ] =
+    useState(initialPlaceholder)
   const [ createdOptions, setCreatedOptions ] = useState<Option[]>([])
+  const [ createNewOption, setCreateNewOption ] = useState<boolean>(false)
+  const [ addFieldInputValue, setAddFieldInputValue ] = useState<string>('')
+
+  const toggleCreateNewOption = () => {
+    setCreateNewOption(not)
+  }
```
olchyk98, line 112: Let's not toggle. Toggling here is quite dangerous and may break if a small change is introduced. Manually setting state value to true/false will give you more control.

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -1,12 +1,13 @@
-import React, { useMemo, useRef, useState } from 'react'
-import { ActionMeta, CreatableSelect } from 'chakra-react-select'
-import type { Option, SingleValue } from '@northlight/ui'
+import React, { useRef, useState } from 'react'
+import { ActionMeta, CreatableSelect, SingleValue } from 'chakra-react-select'
 import { PlusSolid } from '@northlight/icons'
-import { Box, Icon, useOutsideClick } from '@chakra-ui/react'
-import { isNil } from 'ramda'
+import { Box, Icon } from '@chakra-ui/react'
+import R, { isEmpty, not, trim } from 'ramda'
```
olchyk98, line 5: In most cases, we don't import `R` namespace. Try to utilize direct imports only,
such as `import { isEmpty, not, trim }` from 'ramda'

olchyk98 (changes requested):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -78,95 +79,98 @@ import { customComponents } from './custom-components'
...
-  const handleCreateOption = () => {
-    const newOptionTextLower = newOptionText.toLowerCase()
-    const newOptionExists = standardOptions.some((option) =>
-      option.value.toLowerCase() === newOptionTextLower)
+    const newOptionTextLower = R.toLower(newOptionTextString)
 
-    if (!newOptionExists) {
-      const newOption = { label: newOptionText, value: newOptionText }
-      setCreatedOptions((currentCreatedOptions) => [ ...currentCreatedOptions, newOption ])
+    const newOptionExistsInStandardOptions = standardOptions.some(
```
olchyk98, line 129: use `any` from ramda. We try to utilize ramda as much as possible, and mostly not rely on native methods.

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -78,95 +79,98 @@ import { customComponents } from './custom-components'
...
+export const CreatableSelectDropdown = <T extends string = string>({
   standardOptions,
   initialPlaceholder = 'Select or create...',
-  addOptionPlaceholder = 'Enter text...',
+  addOptionPlaceholder = 'Select or create...',
   creationOption = {
-    label: 'Add option...',
-    value: 'Add option...',
+    label: 'Add field',
+    value: 'Add field',
```
olchyk98, line 88: const `CREATION_OPTION_VALUE = 'add_field'`

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -78,95 +79,98 @@ import { customComponents } from './custom-components'
...
     }
-    setSelectedOption(option)
-    onOptionChange(option)
-  }
-
-  const handleKeyDown = (event: { key: string, preventDefault: () => void }) => {
-    if (event.key === 'Enter' && newOptionText) {
-      handleCreateOption()
-      event.preventDefault()
+    if (option.value !== 'Add field') {
```
olchyk98, line 170: use the constant here :)

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -180,36 +184,59 @@ export const CreatableSelectDropdown = <T extends string = string> ({
...
+          onSubmit={ (v) => {
+            handleCreateOption(v)
+          } }
+          variant="brand"
+          sx={ {
+            minWidth: '100%',
+            width: '100%',
+          } }
+        />
+      ) : (
```
TheLudd, line 209: We want to avoid the ternary operator in jsx code. Love it otherwise but rendering components like makes it a bit unreadable.
So we prefer

```
{ condition && (
  <ComponentIfTrue />
) }
{ !condition && (
  <ComponentIfFalse />
) }
```

framework/lib/components/createable-select-dropdown/custom-components.tsx
```diff
@@ -4,20 +4,40 @@ import { PlusSolid } from '@northlight/icons'
...
-      </>
-    </chakraComponents.Option>
+    <Box
+      onClick={ () => {
+        if (props.data.isCreation) {
+          props.selectProps.toggleCreateNewOption()
+        }
+      } }
+    >
+      <chakraComponents.Option { ...props }>
```
TheLudd, line 18: Should it be namespaced?

SebCodesTheWeb:
https://github.com/mediatool/northlight/assets/90923099/8ca204be-776c-4e35-8a7b-e5d492591b3b

Idk what changed, but when I try to click on the add field the UI looks a bit weird, it should toogleCreateOption immediately and display the editable text field

olchyk98 (changes requested):

framework/lib/components/createable-select-dropdown/custom-components.tsx
```diff
@@ -4,20 +4,40 @@ import { PlusSolid } from '@northlight/icons'
...
+            <Flex
+              mr={ 3 }
+              width={ 1.5 }
+              mb={ 0.5 }
+              justifyContent="center"
+              alignItems="center"
+            >
+              <Icon as={ PlusSolid } color="brand" />
+            </Flex>
+          ) : (
```
TheLudd, line 30: Same here with the ternary

TheLudd (changes requested):

framework/lib/components/createable-select-dropdown/custom-components.tsx
```diff
@@ -4,20 +4,40 @@ import { PlusSolid } from '@northlight/icons'
...
+            <Box mr={ 3 } width={ 1.5 } />
+          ) }
```
olchyk98, line 32: we try to not use ternaries in JSX.
we usually go for something like this:
```
{ props.data.isCreation && ... }
{ !props.data.isCreation && ... }
```

olchyk98 (changes requested):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -78,95 +79,98 @@ import { customComponents } from './custom-components'
...
     newValue: SingleValue<Option>,
-    _actionMeta: ActionMeta<Option>) => {
+    _actionMeta: ActionMeta<Option>
+  ) => {
     const option = newValue
 
-    if (option === null) {
+    if (newValue?.isCreation) {
+      toggleCreateNewOption()
+    }
```
SebCodesTheWeb, line 159: @GustafGronlund I think these lines of code are the culprit to the problem in my previous comment. When you click on the 'Add field' it triggers toggleCreateNewOption in custom-components.tsx, but since you also call it in handleChange it ends up being called twice.

SebCodesTheWeb (changes requested):

TheLudd (approved):

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -78,95 +79,98 @@ import { customComponents } from './custom-components'
...
+      standardOptions
+    )
 
-    if (!newOptionExists) {
-      const newOption = { label: newOptionText, value: newOptionText }
-      setCreatedOptions((currentCreatedOptions) => [ ...currentCreatedOptions, newOption ])
+    const newOptionExistsInCreatedOptions = any(
+      (option) => toLower(option.value) === newOptionTextLower,
+      createdOptions
+    )
```
olchyk98, line 134: last comment. remove duplication by moving "any" functionality to a separate function that accepts `standardOptions/createdOptions`.

olchyk98 (changes requested):

olchyk98 (approved):

olchyk98:
@SebCodesTheWeb

PR #272, opened 2024-04-03
fix(framework): fix(framework): fix render and props in dropdown

GustafGronlund:
I've enhanced the CreatableSelectDropdown component by introducing support for restProps. This allows it to accept additional properties, such as 'isDisabled'. Furthermore, I've implemented a useEffect hook that activates when the 'value' prop changes, addressing the previous rendering issues.

### DIFF
--- a/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
+++ b/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react'
+import React, { useEffect, useMemo, useRef, useState } from 'react'
 import { ActionMeta, CreatableSelect, SingleValue } from 'chakra-react-select'
 import { PlusSolid } from '@northlight/icons'
 import { Box, Icon } from '@chakra-ui/react'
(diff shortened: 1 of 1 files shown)


framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -187,6 +188,20 @@ export const CreatableSelectDropdown = <T extends string = string>({
...
+      const newSelectedOption = standardOptions.find((option) => option.value === value)
+        ?? createdOptions.find((option) => option.value === value)
+        ?? null
+
+      if (selectedOption?.value !== newSelectedOption?.value) {
+        setSelectedOption(newSelectedOption)
+      }
+    }
+
+    updateSelectedOption()
```
olchyk98, line 202: why does it have to be in a function that you instantly call?

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -187,6 +188,20 @@ export const CreatableSelectDropdown = <T extends string = string>({
     ...combinedOptions,
   ]
 
+  useEffect(() => {
+    const updateSelectedOption = () => {
+      const newSelectedOption = standardOptions.find((option) => option.value === value)
+        ?? createdOptions.find((option) => option.value === value)
```
olchyk98, line 194: you can use `combinedOptions` here :)

olchyk98 (changes requested):

PR #274, opened 2024-04-05
fix(framework): make zIndex prop optional

SebCodesTheWeb:


framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -60,7 +60,7 @@ export const Toolbox = ({
   resizeLimit = 'full',
   zIndex = coreZIndex.overlay,
   ...rest
-}: Omit<ToolboxProps, 'zIndex'> & { zIndex: number }) => {
+}: Omit<ToolboxProps, 'zIndex'> & { zIndex?: number }) => {
```
olchyk98, line 63: can't we do this `Omit` operation in `ToolboxProps`?
Doing this here increases the chance of devs being confused by ToolboxProps type declarations.

olchyk98 (changes requested):

olchyk98 (approved):

PR #280, opened 2024-04-26
fix(framework): fix width and duplication issue in dropdown

GustafGronlund:
This commit corrects the dropdown width to use full space and removes duplications when creating a new option.

closes: DEV-12853

### DIFF
--- a/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
+++ b/framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
@@ -8,7 +8,8 @@ import {
 } from 'chakra-react-select'
 import { PlusSolid } from '@northlight/icons'
 import { Box, Icon } from '@chakra-ui/react'
-import { any, isEmpty, toLower, trim } from 'ramda'
+import { any, isEmpty, prop, toLower, trim } from 'ramda'
+import { uniqBy } from 'yafu'
 import { customSelectStyles } from '../../theme/components/select/custom-select'
 import type {
   CreatableSelectDropdownProps,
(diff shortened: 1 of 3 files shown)


framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -190,7 +191,7 @@ export const CreatableSelectDropdown = <T extends string = string>({
   }
 
   const combinedOptions = useMemo(
-    (): Option<T>[] => [ ...standardOptions, ...createdOptions ],
+    (): Option<T>[] => uniqBy(compose(toLower, trim, prop('value')), [ ...standardOptions, ...createdOptions ]),
```
SebCodesTheWeb, line 194: Are you sure about `toLower` and `trim`? If there existed an option called 'sweden' and the user then tried to create 'SWEDEN' it would not go through, similiar thing for trim

framework/lib/components/createable-select-dropdown/createable-select-dropdown.tsx
```diff
@@ -8,7 +8,8 @@ import {
 } from 'chakra-react-select'
 import { PlusSolid } from '@northlight/icons'
 import { Box, Icon } from '@chakra-ui/react'
-import { any, isEmpty, toLower, trim } from 'ramda'
+import { any, compose, isEmpty, prop, toLower, trim } from 'ramda'
+import { uniqBy } from 'yafu'
```
SebCodesTheWeb, line 12: you can get uniqBy from ramda aswell, not that it is functionally different, but the bundle size would be smaller since we already have ramda but not yafu as dependency

SebCodesTheWeb: nvm on this comment, Oles enlightened me that it's not available in the version of ramda that we are using

SebCodesTheWeb (approved):

PR #286, opened 2024-05-31
fix(northlight-ui): make color clearable in colorpicker

GustavHF73:
Added a button to the color picker that clears the selected color by setting the color to null, which is default. This makes it possible for the user to revert a color selection to default.

DEV-13921


framework/lib/components/color-picker/color-picker.tsx
```diff
@@ -136,8 +136,22 @@ export const ColorPicker = ({
...
+              alignItems="center"
+              justifyContent="space-between"
+              width="full"
             >
               Pick a color
+              { trigger.color !== 'unset' && (
+                <IconButton
+                  aria-label="Clear color"
+                  size="sm"
+                  height="95%"
```
SebCodesTheWeb, line 149: Why 95%? :)

framework/lib/components/color-picker/color-picker.tsx
```diff
@@ -136,8 +136,22 @@ export const ColorPicker = ({
...
+                  onClick={ () => handleSelect(null) }
+                  variant="ghost"
+                  icon={ <Icon as={ IRemoveSolid } /> }
```
SebCodesTheWeb, line 152: Also I'm a bit unsure about the x, it gives the impression of closing the modal as opposed to reseting selected color. How about a button that says reset?

framework/lib/components/color-picker/color-picker.tsx
```diff
@@ -136,8 +143,22 @@ export const ColorPicker = ({
...
+              display="flex"
+              alignItems="center"
+              justifyContent="space-between"
+              width="full"
             >
               Pick a color
+              { clearable && trigger.color !== 'unset' && (
+                <Button
+                  size="sm"
+                  height="20px"
```
SebCodesTheWeb, line 155: h='5', tokens please :)

SebCodesTheWeb (approved):
I put a small stylistic comment but it's good to go

PR #293, opened 2024-07-03
fix(framework): add cursor pointer to select component

GustavHF73:
There is currently no indication that the select is clickable apart form the caret. This commit adds a cursor pointer to the box element containing the select to clearly show that it is clickable.

DEV-14454


SebCodesTheWeb:
@ilija I will approve this PR but leave the decision up to you

SebCodesTheWeb (approved):

PR #297, opened 2024-07-10
feature(framework): make file picker height customizable

laurabouchara:
closes DEV-14586

### DIFF
--- a/framework/lib/components/file-picker/file-picker.tsx
+++ b/framework/lib/components/file-picker/file-picker.tsx
@@ -60,6 +60,7 @@ export const FilePicker = ({
   loadWithSpinner = false,
   confirmDelete = false,
   editFileOptions = defaultEditOptions,
+  sx = {},
   ...rest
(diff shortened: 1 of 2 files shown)


framework/lib/components/file-picker/file-picker.tsx
```diff
@@ -205,7 +205,11 @@ export const FilePicker = ({
         onMouseEnter={ () => setEditable(true) }
         onMouseLeave={ () => setEditable(false) }
       >
-        <Box sx={ filePicker }>
+        <Box sx={ {
+          ...filePicker,
+          height: rest?.height ?? filePicker.height,
+        } }
+        >
```
SebCodesTheWeb, line 212: I think it would be nice if this worked for other properties as well and not just height. Perhaps we could merge the sx prop with the default filePicker styles like it is done in `drop-zone.tsx` and `theme/components/drag-and-drop/index.ts`

SebCodesTheWeb (approved):

PR #301, opened 2024-07-11
chore(norhtlight): badge sizes improvement

ilija-:
Description
This PR enhances the Badge component by introducing size variants and ensuring that the colorScheme property is correctly typed. The key changes include:

1. **Size Variants:** Added support for sm, md, and lg sizes, allowing more flexibility in the badge's appearance.
2. **ColorScheme Typing:** Utilized the ColorScheme type from @chakra-ui/react to ensure type safety and autocompletion for the colorScheme property.
3. **TypeScript Definitions**: Updated the BadgeProps interface to include the new size and variant properties while extending from ChakraBadgeProps.

These improvements enhance the flexibility and robustness of the Badge component, making it easier to use and maintain.

Changes
1. Updated Badge component to support **2xs,** **xs**(default one that we use right now across the app), **sm** and **md** sizes.
2. Ensured colorScheme property uses ColorSchemes 1:1 from Figma's sketches for the Badge.
3. Extended BadgeProps interface for better type safety and developer experience.

Screenshot 2024-07-11 at 16 42 15

### DIFF
--- a/framework/lib/components/badge/types.ts
+++ b/framework/lib/components/badge/types.ts
@@ -1,0 +1,11 @@
+import { BadgeProps as ChakraBadgeProps, ThemingProps } from '@chakra-ui/react'
+
+type BadgeSize = 'xs' | 'sm' | 'md' | 'lg'
+type BadgeVariant = 'solid' | 'outline' | 'subtle'
+type ColorScheme = ThemingProps['colorScheme']
+
+export interface BadgeProps extends Omit<ChakraBadgeProps, 'size' | 'variant' > {
+  size?: BadgeSize
+  variant?: BadgeVariant
+  colorScheme?: ColorScheme
+}
--- a/framework/lib/theme/components/badge/index.ts
+++ b/framework/lib/theme/components/badge/index.ts
@@ -37,4 +37,21 @@ export const Badge: ComponentSingleStyleConfig = {
       }
     },
   },
+  sizes: {
+    xs: {
+      fontSize: '2xs',
+    },
+    sm: {
+      fontSize: 'xs',
+    },
+    md: {
+      fontSize: 'sm',
+    },
+    lg: {
+      fontSize: 'md',
+    },
+  },
+  defaultProps: {
+    size: 'sm',
+  },
 }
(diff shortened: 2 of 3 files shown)


framework/lib/components/badge/types.ts
```diff
@@ -0,0 +1,11 @@
+import { BadgeProps as ChakraBadgeProps } from '@chakra-ui/react'
+
+type BadgeSize = '2xs' | 'xs' | 'sm' | 'md'
+type BadgeVariant = 'solid' | 'outline' | 'subtle'
+type BadgeColorSchemes = 'gray' | 'mediatoolBlue' | 'blue' | 'red' | 'green' | 'orange' | 'yellow' | 'teal' | 'purple' | 'pink'
```
SebCodesTheWeb, line 5: We already have a type called `Color` in framework/lib/types/color. It does miss the 'teal' color tho so you could add it. A better solution might be to write something like this: `import { palette } from '@northlight/tokens'
const colorSchemes = keys(omit([ 'mono' ], palette))
type ColorScheme = typeof colorSchemes[number]
` - this way the ColorScheme type updates with the tokens

framework/lib/theme/components/badge/index.ts
```diff
@@ -37,4 +37,21 @@ export const Badge: ComponentSingleStyleConfig = {
...
+    sm: {
+      fontSize: 'sm',
+    },
+    md: {
+      fontSize: 'md',
+    },
+  },
+  defaultProps: {
+    size: 'xs',
+  },
```
SebCodesTheWeb, line 56: It seems like at some parts of the mediatool codebase size Badge is used with size="sm", even tho it doesn't make a difference. But if we merge these changes then those badges will suddenly become one size bigger. So the defaultSize should probably be sm instead of xs and everything should move up one step: [2xs, xs, sm, md] => [xs, sm, md, lg].

SebCodesTheWeb (changes requested):
This commit/PR should prob be marked as breaking because it changes the types of BadgeSize and BadgeColorScheme which would deprecate support for some previous props. Also, at some places in the app colorScheme="blackAlpha", but this colorScheme does not seem to do anything, so maybe we should just remove those from app.

SebCodesTheWeb (approved):

PR #302, opened 2024-07-12
chore(northlight): Add semantic colors and cons padding to StatusPin

marcbiemer:
Updated innerPin size for lg to be 4 instead of 3.5 to improve consistency of padding and scaling between size variants. Adjusted colorVariant mapping to use semantic tokens instead of green.500 etc and set bg to transparent. 
Updated documentation by adding more details, linking to related badge component and improving the usage presentation.

Closes: DEV-14603

### DIFF
--- a/framework/lib/components/status-pin/status-pin.tsx
+++ b/framework/lib/components/status-pin/status-pin.tsx
@@ -5,30 +5,35 @@ import { pinSizeMap } from './pin-size-map'
 import { StatusPinProps } from './types'
 
 /**
+ * Status pins are meant to display the status of a specific entry.
+ * @see Badge
  * @see {@link https://northlight/reference/status-pin}
  *
- * @example
+ * @example (Example)
  * (?
  * +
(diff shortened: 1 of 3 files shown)


framework/lib/components/status-pin/status-pin.tsx
```diff
@@ -37,7 +42,7 @@ export const StatusPin = ({ size = 'md', variant, ...rest }: StatusPinProps) =>
   return (
     <Circle
       size={ outerSize }
-      bg="white"
+      bg="transparent"
       borderWidth="1px"
```
ilija-, line 46: borderWidth token here maybe.

ilija- (changes requested):

framework/lib/components/status-pin/status-pin.tsx
marcbiemer: Fixed it :)

ilija- (approved):
GG

PR #309, opened 2024-07-22
feat(northlight-ui): add next week to date range picker

GustafGronlund:
This commit resolves the issue where we wanted to add Next Week as an option in the date range picker. Previously, we had This Week and Last Week, but users can now also select Next Week. 

https://github.com/user-attachments/assets/4ee6ecce-1280-480a-87c5-3a4dd24ae654

closes: #DEV-13744

### DIFF
--- a/framework/lib/components/date-picker/components/calendar/quick-navigation/get-quick-select-options.ts
+++ b/framework/lib/components/date-picker/components/calendar/quick-navigation/get-quick-select-options.ts
@@ -42,6 +42,14 @@ export const getQuickSelectOptions = (
     label: 'This Week',
   }
 
+  const nextWeek = {
+    value: {
+      start: startOfWeek(thisDay.add({ weeks: 1 }), locale),
+      end: endOfWeek(thisDay.add({ weeks: 1 }), locale),
+    },
+    label: 'Next Week',
+  }
+
   const lastWeek = {
     value: {
       start: startOfWeek(thisDay.subtract({ weeks: 1 }), locale),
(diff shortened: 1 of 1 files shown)


TautvydasGec (commented):
can we also sort this to make some sense?
@ilija- or the sort is made specific. Cause its weird that we jump around from month to year then to month again

framework/lib/components/date-picker/components/calendar/quick-navigation/get-quick-select-options.ts
```diff
@@ -264,5 +297,7 @@ export const getQuickSelectOptions = (
     ...fiscalYears,
   ]
 
-  return { quickDates, fiscalQuarters }
+  const sortedQuickDates = sortQuickDatesByLabel(quickDates)
```
TautvydasGec, line 300: do we need to sort them? or could we just change the sort in quick dates array?

TautvydasGec (changes requested):

TautvydasGec (approved):

framework/lib/components/date-picker/components/calendar/quick-navigation/get-quick-select-options.ts
```diff
@@ -250,17 +258,18 @@ export const getQuickSelectOptions = (
...
     nextThreeMonths,
     nextSixMonths,
+    nextMonth,
+    thisMonth,
+    lastYear,
     nextYear,
+    thisYear,
+    yearToDate,
     ...fiscalYears,
   ]
```
SebCodesTheWeb, line 274: Has this order been approved by Kate?

SebCodesTheWeb (approved):

framework/lib/components/date-picker/components/calendar/quick-navigation/get-quick-select-options.ts
GustafGronlund: I talked with Kate this morning and she approved the sorting 🙂

PR #311, opened 2024-07-29
chore(framework): enable override for toolbox body

GustavHF73:
We are accepting sx as an argument to the toolbox component in order to be able to override props in the body

### DIFF
--- a/framework/lib/theme/components/toolbox/index.ts
+++ b/framework/lib/theme/components/toolbox/index.ts
@@ -1,8 +1,9 @@
 import { ComponentMultiStyleConfig } from '@chakra-ui/react'
+import { merge } from 'ramda'
 
 export const Toolbox: ComponentMultiStyleConfig = {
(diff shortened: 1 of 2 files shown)


framework/lib/theme/components/toolbox/index.ts
```diff
@@ -31,14 +32,14 @@ export const Toolbox: ComponentMultiStyleConfig = {
...
+    body: merge({
       p: sizing['4'],
       flexDirection: 'column',
       w: '100%',
       h: '100%',
       maxH: `calc(100vh - ${sizing['16']} * 2)`,
       overflowY: 'scroll',
-    },
+    }, sx.body),
   }),
```
SebCodesTheWeb, line 43: Screenshot 2024-07-29 at 15 29 59
If sx is an empty object then this is gonna crash, I think I would do something like the pic instead
Screenshot 2024-07-29 at 15 36 54

PR #312, opened 2024-07-29
chore(framework): add sx to toolbox content

GustavHF73:
Toolbox content accepts sx as prop in order to be able to override styling. Changed children wrapper in toolbox from box to vstack to be able to use flex grow for children

### DIFF
--- a/framework/lib/components/toolbox/toolbox.tsx
+++ b/framework/lib/components/toolbox/toolbox.tsx
@@ -10,6 +10,7 @@ import { Portal } from '../portal'
 import { ToolboxProps } from './types'
 import { getChildrenWithProps } from '../../utils'
 import { ResizeHandle } from '../resize-handle'
+import { VStack } from '../stack'
 
 /**
  * Controllable Sidebar drawer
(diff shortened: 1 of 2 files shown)


framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import { useMultiStyleConfig, useToken } from '@chakra-ui/system'
 import { coreZIndex } from '@northlight/tokens'
 import { FocusScope } from '@react-aria/focus'
+import { VStack } from '@chakra-ui/react'
```
SebCodesTheWeb, line 5: Screenshot 2024-08-02 at 11 14 45
it should be imported from lib/stack

framework/lib/components/toolbox/toolbox.tsx
```diff
@@ -117,9 +118,9 @@ export const Toolbox = ({
           >
             { isResizable && <ResizeHandle { ...resizeProps } /> }
             <FocusScope autoFocus={ autoFocus }>
-              <Box width="full" height="full">
+              <VStack width="full" height="full">
```
SebCodesTheWeb, line 121: also why this change

GustavHF73: This is because we want the display flex and column on the container, in order to make the flex-grow prop on the children work

SebCodesTheWeb (approved):

PR #315, opened 2024-09-09
fix(framework): hide week numbers if firstDayOfweek is sunday

SebCodesTheWeb:
Having iso Week numbers do not make sense if the weeks themselves are not iso (not starting on monday)
closes: DEV-12024

### DIFF
--- a/framework/lib/components/date-picker/components/calendar/components/standalone-calendar-grid.tsx
+++ b/framework/lib/components/date-picker/components/calendar/components/standalone-calendar-grid.tsx
@@ -69,7 +69,9 @@ export const StandaloneCalendarGrid = ({
         <chakra.table { ...gridProps } sx={ { borderSpacing: '0' } }>
           <chakra.thead { ...headerProps }>
             <chakra.tr>
+              { firstDayOfWeek === 'monday' && (
               <chakra.th />
+              ) }
               <DayLabels weekDays={ weekDays } />
             </chakra.tr>
           </chakra.thead>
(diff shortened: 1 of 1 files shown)


KateKPV:
This needs to be considered @SebCodesTheWeb 
In Sweden, the week numbers follow the international standard ISO-8601, where the week starts on a Monday. The ISO-8601 also stipulates that the first week of the year is the week that contains the first Thursday of the year. This standardization means that if a calendar starts the week on Sunday (often seen in the US), the week numbers could differ from the ISO-8601 system used in Sweden and other parts of Europe.

For example, if January 1st falls on a Sunday:
The ISO-8601 standard will consider the Monday following January 1st as the beginning of week 1.
A Sunday-starting week calendar might mark this Sunday as the start of week 1 instead.
Therefore, the week numbers may not align across these differing systems, which is important to consider for scheduling and planning across different calendars.

magnusohlin:
@KateKPV So are you saying we should hide week numbers all together?

KateKPV:
@magnusohlin no not at all, I think we should support both week starts on Sunday and Monday

magnusohlin:
@KateKPV Understood! So due to the time restrictions that we are under now, would you rather we remove the faulty weeks for the Sunday alternative or do nothing for now?

SebCodesTheWeb:
@KateKPV @magnusohlin, this PR is my suggestion to remove faulty week numbers for now. I have a rough idea for how maybe this could be fixed in the future according to this sketch
F1FB7350-74E9-4C02-8851-E1D67DA29C26_4_5005_c alto there could be a lot of edge cases and take some time to fix this. Problem is that it does not make sense to have ISO week numbers when the week itself is not ISO, we did not have week numbers in the previous date range picker that supported sunday as first day.

PR #322, opened 2024-10-08
fix(framework): added defaultToZeroIfEmpty prop

richardilleskosmala:
With defaultToZeroIfEmpty we can make sure the ComboPicker always has an input value, for the cases in which we don't want input value to be optional

Closes: 86963qkcg

### DIFF
--- a/framework/lib/components/combo-picker/combo-picker.tsx
+++ b/framework/lib/components/combo-picker/combo-picker.tsx
@@ -1,10 +1,11 @@
 import React, { useEffect, useRef, useState } from 'react'
 import { Button, InputGroup, InputRightElement, useDisclosure } from '@chakra-ui/react'
-import { NumberFormatValues } from 'react-number-format'
+import { is } from 'ramda'
 import { ComboPickerOption, ComboPickerProps, ComboPickerValue } from './types'
 import { Select, SingleValue } from '../select'
 import { Box } from '../box'
 import { FormattedNumberInput } from '../text-field'
+import { clamp } from '../../utils'
 
 /**
  * @see ComboPickerField
@@ -14,30 +15,40 @@ export const ComboPicker = ({
   onChange,
   options,
   size,
-  value,
+  value: valueProp,
   placeholder,
   precision,
   formatPreset,
   isDisabled,
   isReadOnly,
+  defaultToZeroIfEmpty = true,
+  max = Infinity,
+  min = -Infinity,
   ...rest
 }: ComboPickerProps) => {
   const { isOpen, onToggle, onClose } = useDisclosure()
-  const [ inputValue, setInputValue ] = useState(value?.input)
-  const [ selectOption, setSelectOption ] = useState(value?.option ?? options[0])
+  const [ inputValueState, setInputValueState ] = useState(valueProp?.input)
+  const [ selectOptionState, setselectOptionState ] = useState(valueProp?.option ?? options[0])
   const [ enableSelectInput, setEnableSelectInput ] = useState(false)
+
+  const isInputValueControlled = typeof valueProp?.input !== 'undefined'
+  const inputValue = isInputValueControlled ? valueProp.input : inputValueState
+
+  const isOptionControlled = typeof valueProp?.option !== 'undefined'
+  const selectOption = isOptionControlled ? valueProp.option : selectOptionState
 
   const buttonRef = useRef<any>()
   const selectRef = useRef<any>()
 
-  const getNewValue = (option: ComboPickerOption, input?: number): ComboPickerValue => ((input)
-    ? { input: Number(input), option }
-    : { option })
+  const getNewValue = (option: ComboPickerOption, input?: number): ComboPickerValue => (
+    (is(Number, input))
+      ? { input: Number(input), option }
+      : { option })
 
-  const handleInputChange = (newInputvalue: NumberFormatValues) => {
-    const newValue = getNewValue(selectOption, newInputvalue.floatValue)
+  const handleInputChange = (newInputvalue?: number) => {
+    const newValue = getNewValue(selectOption, newInputvalue)
 
-    setInputValue(newValue.input)
+    setInputValueState(newValue.input)
     onChange?.(newValue)
   }
 
@@ -52,10 +63,12 @@ export const ComboPicker = ({
 
   const handleSelectChange = (selectedOption: SingleValue<ComboPickerOption>) => {
     if (selectedOption) {
-      setSelectOption(selectedOption)
+      setselectOptionState(selectedOption)
       onChange?.(getNewValue(selectedOption, inputValue))
 
-      handleSelectClose()
+      if (isOpen) {
+        handleSelectClose()
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


framework/lib/components/text-field/formatted-number-input.tsx
```diff
@@ -75,28 +76,41 @@ export const FormattedNumberInput = ({
...
   max = Infinity,
   min = -Infinity,
   inputLeftElement,
   inputRightElement,
+  onBlur,
   ...rest
 }: FormattedNumberInputProps) => {
+  const [ v, setV ] = useState(valueProp)
+  const isControlled = typeof valueProp !== 'undefined'
+  const value = isControlled ? valueProp : v
```
SebCodesTheWeb, line 90: Okay so I'm very nitpicky here and I know the internal state was called "v" before but could you rename it to valueState? So the distinction between valueState and valueProp is clear

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -30,15 +34,24 @@ export const ComboPicker = ({
...
+    (is(Number, input))
+      ? { input: Number(input), option }
+      : { option })
+
+  const getInputWithinRange = ((input: number) => {
+    if (max && input > max) return max
+    if (min && input < min) return min
+
+    return input
+  })
```
SebCodesTheWeb, line 47: Default max=Infinity and min=-Infinity and then return clamp(min, max, input), it's a util in northlight

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -78,13 +91,20 @@ export const ComboPicker = ({
       <FormattedNumberInput
         width="100%"
         onChange={ handleInputChange }
-        defaultValue={ inputValue }
+        value={ inputValue }
```
SebCodesTheWeb, line 94: Hey what happens if I try to change the value prop from outside, would it react and update the internal state of combo-picker?

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -78,13 +91,20 @@ export const ComboPicker = ({
...
         numberOfDecimals={ precision }
         preset={ formatPreset }
         disabled={ isDisabled }
         readOnly={ isReadOnly }
+        min={ min }
+        max={ max }
+        onBlur={ () => {
+          if (!is(Number, inputValue) && defaultToZeroIfEmpty) {
+            setInputValue(0)
+          }
```
SebCodesTheWeb, line 106: Check if inputValue isNil from ramda instead, it should be more direct?

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -30,15 +34,24 @@ export const ComboPicker = ({
...
+    return input
+  })
 
   const handleInputChange = (newInputvalue: NumberFormatValues) => {
     const newValue = getNewValue(selectOption, newInputvalue.floatValue)
 
     setInputValue(newValue.input)
-    onChange?.(newValue)
+    onChange?.({ ...newValue,
+      ...!is(Number, newValue.input) && defaultToZeroIfEmpty && { input: getInputWithinRange(0) } })
```
SebCodesTheWeb, line 54: Maybe put this in a separate variable? `const clampedDefaultValue = ...`

SebCodesTheWeb (changes requested):

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: If the inputValue would be NaN for some reason, it would not be properly handled by isNil, so for all the checks on this component and MoneyInputField I'm just checking if it's a number instead, and if it's not a number than it's not valid

framework/lib/components/text-field/formatted-number-input.tsx
richardilleskosmala: Done!

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: Changed

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: Done!

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: Fixed

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -14,30 +15,40 @@ export const ComboPicker = ({
...
   ...rest
 }: ComboPickerProps) => {
   const { isOpen, onToggle, onClose } = useDisclosure()
-  const [ inputValue, setInputValue ] = useState(value?.input)
-  const [ selectOption, setSelectOption ] = useState(value?.option ?? options[0])
+  const [ inputValueState, setInputValueState ] = useState(valueProp?.input)
+  const [ selectOptionState, setselectOptionState ] = useState(valueProp?.option ?? options[0])
   const [ enableSelectInput, setEnableSelectInput ] = useState(false)
 
+  const isInputValueControlled = typeof valueProp !== 'undefined'
```
SebCodesTheWeb, line 34: isInputValueControlled and isOptionControlled are the same here?

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -73,18 +86,26 @@ export const ComboPicker = ({
     }
   }, [ enableSelectInput ])
 
+  useEffect(() => {
+    if (!is(Number, inputValue) && defaultToZeroIfEmpty) {
+      handleInputChange(clamp(min, max, 0))
+    }
+  }, [ inputValue ])
```
SebCodesTheWeb, line 93: Hmmm handleInputChange calls setInputValueState, which means that if the value is uncontrolled this could turn into a loop, maybe the useEffect dependency should only be valueProp.input?

framework/lib/components/text-field/formatted-number-input.tsx
```diff
@@ -1,10 +1,11 @@
-import React, { ComponentType, useState } from 'react'
+import React, { ComponentType, useEffect, useState } from 'react'
 import { identity, isNil } from 'ramda'
 import {
   InputAttributes,
   NumberFormatValues,
```
SebCodesTheWeb, line 5: This changes for formatted-number-input could be put in a separate commit

framework/lib/components/text-field/formatted-number-input.tsx
```diff
@@ -75,28 +76,41 @@ export const FormattedNumberInput = ({
...
     const factor = isPercentage ? 100 : 1
     if (vNum * factor > max) {
       const newValue = roundToPrecision(max / factor, numberOfDecimals)
-      setV(newValue)
+      setValueState(newValue)
+      onChange({
+        floatValue: newValue,
+        formattedValue: numericFormatter(newValue.toString(), props),
+        value: newValue.toString(),
+      })
```
SebCodesTheWeb, line 104: Maybe put in separate function, getNumberFormatValueObject or similiar? This is a preference detail tho so feel free to skip

SebCodesTheWeb (changes requested):

SebCodesTheWeb (approved):

PR #323, opened 2024-10-10
feat(framework): add util getConsistentRandomColorFromString

SebCodesTheWeb:


framework/lib/utils/get-consistent-random-color-from-string/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  let hash = 5381
+
+  for (let i = 0; i < str.length; i += 1) {
+    const char = str.charCodeAt(i)
+    hash = (hash << 5) + hash + char
+
+    hash &= hash
+  }
+
+  return (Math.abs(hash) % colors.length) + 1
```
Taffr, line 17: Will this ever generate the 0th index of the colors array?

SebCodesTheWeb: Good observation, I removed the +1

Taffr (approved):
LGTM

PR #325, opened 2024-10-15
Trigger onChange with default value and better behavior

richardilleskosmala:
Now when no value prop is provided and default option/input is assigned this will also trigger the component's onChange

Closes DEV-15917

### DIFF
--- a/framework/lib/components/combo-picker/combo-picker.tsx
+++ b/framework/lib/components/combo-picker/combo-picker.tsx
@@ -28,7 +28,7 @@ export const ComboPicker = ({
 }: ComboPickerProps) => {
   const { isOpen, onToggle, onClose } = useDisclosure()
   const [ inputValueState, setInputValueState ] = useState(valueProp?.input)
-  const [ selectOptionState, setselectOptionState ] = useState(valueProp?.option ?? options[0])
+  const [ selectOptionState, setselectOptionState ] = useState(valueProp?.option)
   const [ enableSelectInput, setEnableSelectInput ] = useState(false)
 
   const isInputValueControlled = typeof valueProp?.input !== 'undefined'
@@ -40,10 +40,13 @@ export const ComboPicker = ({
   const buttonRef = useRef<any>()
   const selectRef = useRef<any>()
 
-  const getNewValue = (option: ComboPickerOption, input?: number): ComboPickerValue => (
-    (is(Number, input))
-      ? { input: Number(input), option }
-      : { option })
+  const getNewValue = (option?: ComboPickerOption, input?: number): ComboPickerValue => {
+    const newValueOption = option ?? options[0]
+
+    return (is(Number, input))
+      ? { input: Number(input), option: newValueOption }
+      : { option: newValueOption }
+  }
 
   const handleInputChange = (newInputvalue?: number) => {
     const newValue = getNewValue(selectOption, newInputvalue)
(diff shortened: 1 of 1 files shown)


framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -28,7 +28,7 @@ export const ComboPicker = ({
 }: ComboPickerProps) => {
   const { isOpen, onToggle, onClose } = useDisclosure()
   const [ inputValueState, setInputValueState ] = useState(valueProp?.input)
-  const [ selectOptionState, setselectOptionState ] = useState(valueProp?.option ?? options[0])
+  const [ selectOptionState, setselectOptionState ] = useState(valueProp?.option)
```
SebCodesTheWeb, line 31: Should this no longer be used as default?

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -87,10 +90,24 @@ export const ComboPicker = ({
   }, [ enableSelectInput ])
 
   useEffect(() => {
-    if (!is(Number, inputValue) && defaultToZeroIfEmpty) {
-      handleInputChange(clamp(min, max, 0))
+    const needsToCorrectOption = !selectOption
```
SebCodesTheWeb, line 93: mby use isNil here, just to be a bit more clear cause javascript type conversions can be a bit confusing

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -87,10 +90,24 @@ export const ComboPicker = ({
...
+    const needsToCorrectInput = !is(Number, inputValue) && defaultToZeroIfEmpty
+
+    const option = needsToCorrectOption ? options[0] : selectOption
```
SebCodesTheWeb, line 96: Does selectedOption now also have a defaultValue? And what about defaultValue prop instead of having options[0] always as the default

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -87,10 +90,24 @@ export const ComboPicker = ({
...
+
+    if (needsToCorrectInput) {
+      setInputValueState(input)
+    }
+
+    if (needsToCorrectOption || needsToCorrectInput) {
+      onChange?.(getNewValue(option, input))
     }
-  }, [ valueProp?.input, defaultToZeroIfEmpty ])
+  }, [ inputValue, defaultToZeroIfEmpty ])
```
SebCodesTheWeb, line 110: Why listen to inputValueState here? Maybe put the clamping in the handleInputChange callback instead, so you won't have to set a state in a useEffect where that state also is a dependency

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: There is the potential case in which NaN could be passed as the inputValue, and that would not be considered nil, that's why I considered checking for Number and exclude any other case

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: Unless I'm misunderstanding your question, we always had a default selectedOption which is options[0], for the case in which the user does not pass one, but now the place where we check if it had a selectedOption when rendered or not is here, and then we assign the default value, it only changed the place where it happens

framework/lib/components/combo-picker/combo-picker.tsx
richardilleskosmala: The way it was before, it would only clamp to 0 if were recieved a new value in valueProp?.input, which is mostly the case of forms, in which when we empty the input, the value in the form will be changed and passed to the moneyInput as 0, but when the component is not inside a form or the valueProp is not controlled, if you empty the input completely it will never enter useEffect to set it to 0

framework/lib/components/combo-picker/combo-picker.tsx
SebCodesTheWeb: I meant for selectOption

framework/lib/components/combo-picker/combo-picker.tsx
SebCodesTheWeb: yes but why remove it from the useState defaultValue

framework/lib/components/combo-picker/combo-picker.tsx
SebCodesTheWeb: can skip this

framework/lib/components/combo-picker/combo-picker.tsx
SebCodesTheWeb: okay so you can skip this aswell

SebCodesTheWeb (changes requested):

SebCodesTheWeb (approved):

PR #341, opened 2024-12-19
fix(northlight-ui): fix weeks in date picker

GustafGronlund:
This commit resolves the issue where the weeks in the date picker were incorrect. The previous solution did not follow the ISO week numbering rules, resulting in inaccurate week numbers. We are now taking into account the actual calendar days instead of simply counting and dividing by 7. 

https://github.com/user-attachments/assets/ffeb469f-4498-4237-acdc-5ce02a2e6a19

closes: #DEV-17786


SebCodesTheWeb:
Screenshot 2024-12-19 at 11 32 22
Screenshot 2024-12-19 at 11 32 31

Something weird seems to be up with the weeknumbers from DEC 2026 - JAN 2027, it seems to miss week 53.

SebCodesTheWeb (approved):

PR #347, opened 2025-02-05
chore(northlight): displayLimit select prop

ilija-:
Introducing new prop `displayLimit`

`displayLimit`: Sets the number of selected items to display before showing the +N indicator.

Uploading Screen Recording 2025-02-05 at 16.21.50.mov…


framework/lib/components/search-bar/get-components.tsx
```diff
@@ -5,59 +5,94 @@ import {
...
+<T extends SearchBarOptionType> = ExtendedProps<T, MenuProps<T, boolean, GroupBase<T>>>
+type ExtendedDropdownIndicatorProps
+<T extends SearchBarOptionType> = ExtendedProps<T, DropdownIndicatorProps<T, boolean, GroupBase<T>>>
+type ExtendedOptionProps
+<T extends SearchBarOptionType> = ExtendedProps<T, OptionProps<T, boolean, GroupBase<T>>>
+type ExtendedMultiValueGenericProps
+<T extends SearchBarOptionType> = ExtendedProps<T, MultiValueGenericProps<T, boolean, GroupBase<T>>>
+type ExtendedControlProps
+<T extends SearchBarOptionType> = ExtendedProps<T, ControlProps<T, boolean, GroupBase<T>>>
+
```
SebCodesTheWeb, line 46: You can add displayLimit?: number in framework/general.d.ts instead. It will then automatically be inferred

framework/lib/components/search-bar/search-bar.tsx
```diff
@@ -43,16 +47,22 @@ export const SearchBar = forwardRef(
       [ debouncedWaitTime ]
     )
     const handleChange = useSelectCallbacks<T, K>({
-      onChange,
+      onChange: onChange as (
+        val: MultiValue<T> | SingleValue<T>,
+        event: ActionMeta<T>
+      ) => void,
```
SebCodesTheWeb, line 53: This shouldn't be necessary

PR #352, opened 2025-03-19
chore(framework): add buttonlabel prop to date range picker

gusekl:
Adds optional prop buttonlabel to date range picker. Defaults to 'Save'.


framework/lib/components/date-picker/components/calendar/types.ts
```diff
@@ -10,6 +10,7 @@ export interface RangeCalendarProps extends AriaRangeCalendarProps<DateValue> {
   isClearable: boolean
   firstDayOfWeek: FirstDayOfWeek
   onSave?: () => void
+  buttonLabel: string
```
olchyk98, line 13: let's make it optional and default to `Save` here as well

olchyk98 (changes requested):

SebCodesTheWeb (commented):
Looks good to me just make buttonLabel prop optional .
Why is this needed tho, feels like we shouldn't tamper with the inner labels of the date range picker?

olchyk98 (approved):

olchyk98:
@SebCodesTheWeb We have a specific case in sketches where we need to change label for submit button to improve clarity in design. This was previously discussed and agreed upon with @HannaYankouskaya.

PR #356, opened 2025-03-31
fix(northlight-ui): update type in flipbuttongroupfield

GustafGronlund:
This commit resolves the issue where the FlipButtonGroupField wouldnt show all the values. This was due to the props not being able to take string[] which can be the case when selecting datasets in reports for example. Ive added the possibility to pass string[] aswell to the value so we can now pass array with string. 

closes: #DEV-19393


SebCodesTheWeb (approved):
nice

PR #361, opened 2025-04-10
fix(framework): added docs for how to use is loading

JacobHellmark:
This is done due to people adding the spinner in an incorecct way

### DIFF
--- a/framework/lib/components/button/button.tsx
+++ b/framework/lib/components/button/button.tsx
@@ -24,6 +24,31 @@ const SpinnerSizeMap = {
         </Button>
  * ?)
  *
+ * @example
+ * ### Use the `isLoading` prop to show a loading spinner
+ * (?
+        <Button
+            variant="success"
+            isLoading={ true }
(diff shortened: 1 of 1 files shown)


framework/lib/components/button/button.tsx
```diff
@@ -24,6 +24,29 @@ const SpinnerSizeMap = {
         </Button>
  * ?)
  *
+ * @example
+ * (?
+        <Button
+            variant="success"
```
SebCodesTheWeb, line 30: Just add some markdown to explain that both of these solutions are valid and when to use them

SebCodesTheWeb (approved):
nice

ilija-:
Can we merge this one ? @JacobHellmark @SebCodesTheWeb

JacobHellmark:
> Can we merge this one ? @JacobHellmark @SebCodesTheWeb

ops thought this was merged @SebCodesTheWeb this should be ok to merge now right?

PR #362, opened 2025-04-29
fix(framework): add white variant for flip button

Emogensen:
New variant added for white background on selection.


ilija- (dismissed):
GG

ilija- (approved):
1. Used existing tokens
2. Adapted the ui according to the figma variants


https://github.com/user-attachments/assets/dfcd6c96-3bc6-4064-8e86-a46628674b72

PR #363, opened 2025-05-06
fix(framework): extends comboPicker to take chakra props

gusekl:
Extends type of comboPicker to accept chakra props

### DIFF
--- a/framework/lib/components/combo-picker/combo-picker.tsx
+++ b/framework/lib/components/combo-picker/combo-picker.tsx
@@ -109,6 +109,8 @@ export const ComboPicker = ({
     }
   }, [ inputValue, defaultToZeroIfEmpty ])
 
+  const buttonWidth = (buttonRef.current?.offsetWidth ?? 0)
+
   return (
     <InputGroup>
       <FormattedNumberInput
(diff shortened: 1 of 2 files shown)


framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -125,22 +125,16 @@ export const ComboPicker = ({
         max={ max }
         { ...rest }
       />
-      <InputRightElement
-        width="auto"
-        display="flex"
-        justifyContent="flex-end"
-        padding={ 2 }
+      <Button
+        ref={ buttonRef }
```
SebCodesTheWeb, line 129: Why we remove this?

gusekl: BC the input gets put behind the button if the inputgroup is not the parent

framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -124,6 +126,7 @@ export const ComboPicker = ({
         min={ min }
         max={ max }
         { ...rest }
+        pr={ `${buttonWidth >= 0 ? buttonWidth + 10 : 0}px` }
```
SebCodesTheWeb, line 129: I would add some comment as to why 10 px specifically, or test/find out if that's the optimal. But I'm nitpicking here so it's ok

PR #368, opened 2025-05-22
feat(northlight): badge icon

ilija-:
This commit adds the possibility to add an icon and iconPosition to a badge.

**Result:**

https://github.com/user-attachments/assets/1fd041a7-47fe-4288-9da7-a83f662491ea

### DIFF
--- a/framework/lib/components/badge/badge.tsx
+++ b/framework/lib/components/badge/badge.tsx
@@ -1,11 +1,16 @@
 import React, { forwardRef } from 'react'
 import { Badge as ChakraBadge } from '@chakra-ui/react'
 import { MediatoolLogoSolid } from '@northlight/icons'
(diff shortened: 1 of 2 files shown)


SebCodesTheWeb (changes requested):
Even if I fancy this addition as a nice prop, I think that it's better to maintain a composable, more generic approach. What I mean is that we instead of adding props to customize <Badge>, withIcon, iconPosition, iconAs, which does not scale well when we in edge cases possibly want to handle things such as alignment, color, other styling etc and have to keep on adding more props. We can instead follow a generic approach where we let the developers customize their icons by writing their own component and styling which is already supported like in screenshot below. 
Screenshot 2025-05-26 at 17 15 21

SebCodesTheWeb (approved):

PR #369, opened 2025-06-25
fix(northlight): alert title and desc

ilija-:
Fixes the issue with the line-height and word break for the title and description.

From this:
Screenshot 2025-06-25 at 15 57 27

To this:
Screenshot 2025-06-25 at 15 44 07
Screenshot 2025-06-25 at 15 42 49


framework/lib/theme/components/alert/index.ts
```diff
@@ -14,6 +14,14 @@ export const Alert: ComponentMultiStyleConfig = {
       width: 'auto',
       display: 'flex',
     },
+    title: {
+      lineHeight: '1.25',
+      wordBreak: 'break-word',
```
KristofferLinderman, line 19: This is technically deprecated, use `overflow-wrap: anywhere ` instead 🤓 
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#syntax

KristofferLinderman (changes requested):
Just a small feedback 🕵️‍♀️

framework/lib/theme/components/alert/index.ts
ilija-: ima webkit boy :) Will amend in a min.

KristofferLinderman (approved):
LGTM ✨

PR #382, opened 2025-07-29
fix(northlight): scrollbar and menu tokens

ilija-:
This commit fixes the docs pages scrollbar and menu refference tokens and replaces them with system tokens

### DIFF
--- a/docs/src/docs/components/chakra-docs-link/index.tsx
+++ b/docs/src/docs/components/chakra-docs-link/index.tsx
@@ -10,7 +10,7 @@ export const ChakraDocsLink = ({ slug, ...rest }: Props) => (
   <Link
     { ...rest }
     target="_blank"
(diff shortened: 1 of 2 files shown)


docs/src/docs/components/chakra-docs-link/index.tsx
```diff
@@ -10,7 +10,7 @@ export const ChakraDocsLink = ({ slug, ...rest }: Props) => (
   <Link
     { ...rest }
     target="_blank"
-    color="blue.500"
+    color="brand"
```
HannaYankouskaya, line 13: Suggestion: text.link.default

ilija-: done ✨

HannaYankouskaya (approved):

PR #394, opened 2025-08-13
fix(northlight-ui): handle undefined theme

JonathanNeville:
before would not handle undefined properly


ilija- (approved):
Looks ok to me. Im just concerned about Path returning any , but ok.

JonathanNeville:
> Looks ok to me. Im just concerned about Path returning any , but ok.

I provided string as a type parameter so it return type should always be string now

PR #395, opened 2025-08-15
fix(framework): fix combopicker to not handle state

gusekl:
Previously combopicker and formatted number input had state handling that led to stale values being displayed. This commit fixes so that the values displayed are the latest values that are passed into the components. The reason for this fix is that handsontable does not remount the editor between cells. This lead to a stale state persisting after editing one cell and going to another.

related to dev-21146

### DIFF
--- a/framework/lib/components/combo-picker/combo-picker.tsx
+++ b/framework/lib/components/combo-picker/combo-picker.tsx
@@ -5,7 +5,6 @@ import { ComboPickerOption, ComboPickerProps, ComboPickerValue } from './types'
 import { Select, SingleValue } from '../select'
 import { Box } from '../box'
 import { FormattedNumberInput } from '../text-field'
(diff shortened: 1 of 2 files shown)


framework/lib/components/combo-picker/combo-picker.tsx
```diff
@@ -90,24 +82,24 @@ export const ComboPicker = ({
...
 
     if (needsToCorrectOption) {
-      setselectOptionState(option)
+      setSelectOption(option)
     }
 
     if (needsToCorrectInput) {
-      setInputValueState(input)
+      setInputValue(input)
     }
```
TautvydasGec, line 97: this code seems weird.
defining const and then setting it. also its unsafe cause both could be undefined and there is not explicit handling for it is there?

PR #407, opened 2025-10-08
fix(framework): add null check to savedDateRange and defaultDateRange

gusekl:
Adds null check to savedDateRange and defaultDateRange. previously they were set to value. This lead to the reset button never showing up if defaultdaterange was not provided.


KristofferLinderman (commented):
Found some iffy thingz:

1. Open the Filter/date thing
2. Open the calendar
3. Close it (apply, click outside etc)
4. Open calendar again:

Image

Should also fix the flickering when resetting:

https://github.com/user-attachments/assets/8d6a0f10-0728-4476-a734-275f27d09270

KristofferLinderman (approved):

PR #456, opened 2026-04-14
fix(northlight): clearable select

ilija-:
Fixes the issue where Clearable `<Select>` is not clearing when clicked on the [X]

**Before:**

https://github.com/user-attachments/assets/cd99d7af-e1b5-4cbc-a8b7-0eee2a54505a

**After:**

https://github.com/user-attachments/assets/bbe63a8e-ce81-4a5e-92f3-bce4c3da0f40


jonglmt (approved):
lgtm

PR #457, opened 2026-04-14
fix(northlight): docs page scroll

ilija-:
Turns out the Page wrappers were basically hijacking scroll. We had `h="100%"` combined with `overflowY="auto"` and `overscrollBehavior="contain"`, which created these isolated scroll containers. Since the global theme already forces body and #app to be 100% height, everything ended up locked to the viewport. On top of that, `overscrollBehavior: contain` made sure scroll events never bubbled up to the window. So yeah, scrolling got stuck.

What I changed:

`reference-page-wrapper.tsx`
Switched `h="100%`" to `minH="100%"` and removed the `overflow` + `overscroll` props from the outer container. Also removed `overflowY="scroll"` from the inner one. Now the content just behaves normally and scrolls with the page.

`page.tsx`
Same cleanup here. Replaced `h="100%"` with `minH="100%"` and removed the `overflow/overscroll` combo from both containers.

`side-nav-bar.tsx`
Dropped` h="full"` since it wasn’t really needed. Also added `pointerEvents="none"` to the fixed container so it stops blocking scroll on the right side. Wrapped links with `pointerEvents="auto"` so they still work.

Thank me later.

**Before:**

https://github.com/user-attachments/assets/bca29ee5-548d-4605-82cb-b65409f66b68

**After:**

https://github.com/user-attachments/assets/c278c727-b28e-4f4a-8706-b41174738643

### DIFF
--- a/docs/src/app/components/page.tsx
+++ b/docs/src/app/components/page.tsx
@@ -14,10 +14,8 @@ export const Page = ({
 }: Props) => (
   <Box
     w="100%"
-    h="100%"
+    minH="100%"
     p={ 6 }
-    overflowY="auto"
-    overscrollBehavior="contain"
     bgColor="background.default"
     color="text.default"
   >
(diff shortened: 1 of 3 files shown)


jonglmt (approved):
finally, thanks for fixing it

PR #458, opened 2026-04-23
fix(framework): fix use-select-callbacks hook

gusekl:
Adds null check to prevent undefined error in use-select-callbacks hook.


KristofferLinderman (approved):
LGTM 🐼✨

PR #459, opened 2026-04-23
fix(framework): add inner horizontal padding to table cells

gusekl:
- Add inner horizontal padding to table cells

Move horizontal padding from table element to first/last cells so that td border-top spans the full width of the table. Previously, px on the table element caused row separator borders to stop short of the outer border, leaving a visible gap at the edges.

### DIFF
--- a/framework/lib/theme/components/table/index.ts
+++ b/framework/lib/theme/components/table/index.ts
@@ -20,7 +20,6 @@ export const Table: ComponentMultiStyleConfig = {
         borderRadius: 'md',
         borderColor: 'border.default',
         borderWidth: 'xs',
-        px: '3',
       },
 
       th: {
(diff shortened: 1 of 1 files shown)


framework/lib/theme/components/table/index.ts
```diff
@@ -20,19 +20,24 @@ export const Table: ComponentMultiStyleConfig = {
...
-        px: '3',
       },
 
       th: {
         color: 'text.subdued',
         fontWeight: 'bold',
         textTransform: 'none',
         fontSize: 'sm',
+        // Add inner horizontal padding (6 from size md + 3 for table inset)
+        '&:first-of-type': { pl: '9' },
```
KristofferLinderman, line 31: Minor implication here is that it now is "too small/big" if you change the `size` on table. But seems to be like 0.5 rem, so might not be noteworthy (if even used)

KristofferLinderman (approved):
Seems to be an somewhat imperfect improvement, but improvement nonetheless!

PR #460, opened 2026-05-15
feat(northlight): DateRangePicker sizes

ilija-:
Adds a size prop to the DateRangePicker according to the Figma sketches.

Figma sketches:
Screenshot 2026-05-15 at 15 45 45

Actual result:

https://github.com/user-attachments/assets/7df26ef1-ccd5-42c0-bea5-f90c11629281

### DIFF
--- a/framework/lib/components/date-picker/components/date-field/trigger.tsx
+++ b/framework/lib/components/date-picker/components/date-field/trigger.tsx
@@ -4,6 +4,24 @@ import { Button } from '../../../button'
 import { TriggerProps } from './types'
 import { Icon } from '../../../icon'
 
+const sizeToButtonSize = {
+  sm: 'xs',
+  md: 'sm',
+  lg: 'md',
+} as const
+
+const sizeToBoxSize = {
+  sm: 6,
+  md: 8,
+  lg: 10,
+}
+
+const sizeToIconBoxSize = {
+  sm: 3,
+  md: 4,
+  lg: 5,
+}
+
 export const Trigger = (props: TriggerProps) => {
   const {
     id,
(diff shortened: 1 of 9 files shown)


framework/lib/components/date-picker/components/date-field/types.ts
```diff
@@ -7,20 +7,24 @@ import { AriaButtonProps } from '@react-aria/button'
 export type DateSegmentType = DateSegment
 export interface DateFieldProps extends AriaDateFieldProps<DateValue> {
   dateFormat?: string
+  size?: 'sm' | 'md' | 'lg'
```
KristofferLinderman: Create a type eg. `DatePickerSize` for this instead of repeating x4

framework/lib/components/date-picker/components/date-field/trigger.tsx
```diff
@@ -4,6 +4,24 @@ import { Button } from '../../../button'
 import { TriggerProps } from './types'
 import { Icon } from '../../../icon'
 
+const sizeToButtonSize = {
+  sm: 'xs',
```
KristofferLinderman, line 8: Create them as type `Record<DatePickerSize, number>` instead so it's always in sync with the avalible sizes

KristofferLinderman (changes requested):
Nice addition to the date range picker 👍 

Things in addition to in-line comments:
1. The commits should be `feat` not `chore`
2. We might as well add support for `size` to the DatePicker as well? It uses the same components as date range picker
3. Could optionally add `size` to the existing (or create new) examples in the docs

ilija-:
> Nice addition to the date range picker 👍
> 
> Things in addition to in-line comments:
> 
> 1. The commits should be `feat` not `chore`
> 2. We might as well add support for `size` to the DatePicker as well? It uses the same components as date range picker
> 3. Could optionally add `size` to the existing (or create new) examples in the docs

1. Done
2. Yeah will do that in a separate pr.
3. yup

KristofferLinderman (approved):
LGTM 🐼✨

PR #461, opened 2026-05-21
chore: add workflow helath check

TheLudd:
This will periodically check if our tokens are valid. The same script is
used in connectors and will warn when a token has expired.


TheLudd:
Will be solved in another PR

PR #462, opened 2026-05-21
Fix nx cache, and add action health check

TheLudd:
I really only wanted the health check but had to fix the cache error due to tests failing.


KristofferLinderman (approved):
LGTM

PR #467, opened 2026-06-17
feat(northlight): trigger release

s1monbengtsson:
Dummy commit to trigger release. Previous commit adding new icons was written as a chore, so did not trigger release to npm


LajnaLegenden (approved):
can you move this to the other side and make this right to left compatible

